PR #86, opened 2022-04-11
[manager]feature: zh and en annotations for code, delete notification when edit query is empty

wang1027-wqh:
Delete notification operation compatibility query is empty
When the delete query is empty, you will be prompted to check the corresponding parameters


tomsun28 (commented):
LGTM Great!

PR #88, opened 2022-04-13
Modify monitoring and add transaction support.

wang1027-wqh:
Modify monitoring and add transaction support.
Monitoring compatible with Chinese and English #wqh


common/src/main/java/com/usthe/common/entity/manager/Param.java
```diff
@@ -33,21 +35,26 @@
 @Builder
 @AllArgsConstructor
 @NoArgsConstructor
-@ApiModel(description = "参数实体")
+@ApiModel(description = "嗯： Parameter Entity,zh: 参数实体")
```
tomsun28, line 38:
```suggestion
@ApiModel(description = "en： Parameter Entity,zh: 参数实体")
```

tomsun28 (commented):
LGTM! 👍👍👍

PR #90, opened 2022-04-14
feat: [manager]feature:App/Account Controller compatible with Chinese and English

wang1027-wqh:
add: App/Account Controller compatible with Chinese and English


wang1027-wqh:
Add: AppController and AccountController compatible with Chinese and English

tomsun28 (commented):
LGTM

PR #101, opened 2022-04-17
[monitor]feature: support roles permission, admin-user-guest

tomsun28:
support roles permission, admin-user-guest
default account admin has [admin] role, can add , edit, view ,delete
default account tom has [user] role, can add, edit, view
default account guest has [guest] role, only can view


tomsun28:
refer to #45 thanks @gcdd1993

PR #247, opened 2022-08-12
[collector] bugfix: Solve imprecise cyclicJob execution interval problem.

cuipiheqiuqiu:
[collector] bugfix: Solve imprecise cyclicJob execution interval problem.
1.wheelTimer tickDuration from 10s changed to 1s,ticksPerWheel remain unchanged.
2.now support second level job interval.


tomsun28 (approved):
LGTM 👍👍👍

PR #299, opened 2022-09-19
[common]feature:use "apache.http.conn.util" replace "sun.net.util" for upgrading java version

Privauto:
[common]feature:use "apache.http.conn.util" replace "sun.net.util" for upgrading java version


tomsun28 (approved):
👍👍 thanks! LGTM.

PR #365, opened 2022-10-17
[web-app] fix history echarts chart tooltip only show one instance

tomsun28:
before: 

now:


wang1027-wqh:
The thread level monitoring and GC frequency monitoring should be performed later

Ceilzcx (approved):
I have tested it in the iot branch and there is no problem

PR #373, opened 2022-10-20
[webapp,warehouse] support historical interval aggregate data chart

tomsun28:


tomsun28:
Interval data chart(max min mean) not supported when has multi instance in this metric.

Ceilzcx (approved):
LGTM

PR #381, opened 2022-10-22
SummaryControllerTest Test class  writing completed

TJxiaobao:
The test class is written and the test is completed


tomsun28 (approved):
Great! 👍👍👍

PR #383, opened 2022-10-23
[collector] fix the snmp service create too many thread but not reused. 

tomsun28:


tomsun28:
This due to the windows monitor use snmp protocol, the snmp service create too many thread but not reused, cause oom.



Fix by pr https://github.com/dromara/hertzbeat/pull/383 

After reused snmp thread.

PR #417, opened 2022-11-06
[collector] support more powerful replace placeholder ^_^ and -_-

tomsun28:


tomsun28:
before: 

we can replace complete field value by -_-filed-_-
eg: The field is name: tom, -_-name-_- can replace tom

now:

we can replace field in value subString eg: xxxx-_-name-_-xxxx can replace xxxxtomxxxx

PR #418, opened 2022-11-06
[collector] support trigger and grading multiple subtasks through -_- placeholder expression

tomsun28:


tomsun28:
Before, we can run the task to obtain the pre-index parameters in a monitoring, and then assign this index as a variable to the next collection task. For example, we first log in with the account and password to obtain the token, and then use the token for the next task to authenticate.

But, In some case, eg: we need to collect multiple docker ids, and then traverse these ids to request the http api to obtain container detail info. The key here is that there are multiple ids, that is, there will be multiple task collections in the next step.

That's what this PR is trying to solve. 

When the pre ids is multi, we trigger multiple subtasks thread to collect it, and merge their response data at the end.

PR #424, opened 2022-11-08
[OS Monitor]bugfix:Fix cpu cores and interrupt acquisition under Orac…

assassinfym:
Under the Oracle Linux based on the arm architecture, there is a problem in collecting cpu cores and interrupt, and it will display&nbsp

image

After analyzing the source code, it is found that it is a script for collecting cpu, and there are some compatibility problems

Original collection script

LANG=C lscpu | awk -F: '/Model name/ {print $2}';awk '/processor/{core++} END{print core}' /proc/cpuinfo;uptime | sed 's/,/ /g' | awk '{for(i=NF-2;i<=NF;i++)print $i }' | xargs;vmstat 1 1 | awk 'NR==3{print $11}';vmstat 1 1 | awk 'NR==3{print $12}';vmstat 1 1 | awk 'NR==3{print $15}'

Where LANG=C lscpu | awk -F: '/Model name/ {print $2}'; , because Oracle Linux will have two nodes, the Model Name and BIOS Model Name, so there are two lines of records when getting info
image

I changed it to

LANG=C lscpu | awk -F: '$1=="Model name" {print $2}';awk '/processor/{core++} END{print core}' /proc/cpuinfo;uptime | sed 's/,/ /g' | awk '{for(i=NF-2;i<=NF;i++)print $i }' | xargs;vmstat 1 1 | awk 'NR==3{print $11}';vmstat 1 1 | awk 'NR==3{print $12}';vmstat 1 1 | awk 'NR==3{print $15}'

The script that gets info LANG=C lscpu | awk -F: '$1=="Model name" {print $2}';
image

### DIFF
--- a/manager/src/main/resources/define/app/app-linux.yml
+++ b/manager/src/main/resources/define/app/app-linux.yml
@@ -114,7 +114,7 @@ metrics:
       username: ^_^username^_^
       password: ^_^password^_^
       timeout: ^_^timeout^_^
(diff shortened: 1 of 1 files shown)


tomsun28 (approved):
LGTM, Thanks!👍👍 Test pass.

PR #515, opened 2022-12-23
[manager] springboot2 monitor support base path config

tomsun28:


tomsun28:
fix issue https://gitee.com/dromara/hertzbeat/issues/I66O68

PR #541, opened 2023-01-04
remove unused host bind ports

liudonghua123:
There is no need to mapping `mysql` and `tdengine` service ports to host. If there is a host `mysql` deamon, the port will be conflict.


tomsun28:
👍👍 Agree, there is no need to expose their ports.  The last docker-compose script is refreshed. Now we have hertzbeat-mysql-iotdb and hertzbeat-mysql-tdengine. Can you modify them with latest code in master branch. https://github.com/dromara/hertzbeat/tree/master/script/docker-compose

liudonghua123:
ok, I will update the pr later.

PR #591, opened 2023-01-17
[manager, web-app] Notification type supports slack webhook, upgrade jdk version to JDK11

gcdd1993:
- upgrade jdk version to JDK11
- upgrade pmd plugin to support jdk11
- [common] fix pmd check
- [manager, web-app] Notification type supports slack webhook


tomsun28 (approved):
Great!👍👍👍

PR #762, opened 2023-03-23
[code]feature:  Customize Redis DB

zqr10159:
Modifying `warehouse.store.redis.db` in `applicationyml` can specify which database Redis should use.


tomsun28 (approved):
thanks!👍👍. 
suggest the `RealTimeRedisDataStorage .saveData` method should config db number when insert data

PR #783, opened 2023-03-27
add springboot2,springboot3 define yml comment translation

topsuder:
add springboot2,springboot3 define yml comment translation


tomsun28:
@all-contributors please add @topsuder  for code

PR #866, opened 2023-04-12
[refactor] use final instead of @Autowired and other code style

gcdd1993:
## What's changed?

使用final替换`@Autowired`，Springboot官方建议使用final来修饰成员变量，然后通过构造方法来进行注入。

## Checklist

- [x]  I have read the [Contributing Guide](https://hertzbeat.com/docs/others/contributing/)


manager/src/main/java/org/dromara/hertzbeat/manager/controller/AppController.java
```diff
@@ -17,37 +17,39 @@
...
 import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
 
 /**
  * Monitoring Type Management API
  * 监控类型管理API
+ *
  * @author tomsun28
  * @date 2021/11/14 16:47
  */
 @Tag(name = "Monitor Type Manage API | 监控类型管理API")
```
rellyrell7898, line 46: - [ ] @********

rellyrell7898 (approved):

gcdd1993:
https://yanbin.blog/mockito-injectmocks-initialize-tested-instance/
其中提到
> 如果存在一个有参数的构造函数，那么 setter 方法 和 属性  注入都不会发生。也就是说 非默认构造函数 不会与后两种方式同时发生，但找不到 setter 注入的 Mock 对象还会尝试用 属性 来直接注入。

当有构造器和`@Autowired`同时存在时，`@InjectMocks`只能注入构造器的，导致测试类无法通过

hertzbeat:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
https://yanbin.blog/mockito-injectmocks-initialize-tested-instance/
which mentioned
> If there is a parameterized constructor, neither setter method nor property injection will happen. That is to say, the non-default constructor will not happen at the same time as the latter two methods, but the Mock object that cannot find the setter injection will try to use the property to inject directly.

When there is a constructor and `@Autowired` exist at the same time, `@InjectMocks` can only inject the constructor, causing the test class to fail

PR #908, opened 2023-04-23
[bug#877]Maximum length control for password increase

hudongdong129:
Maximum length control for password increase


tomsun28 (approved):
LGTM!👍

PR #1636, opened 2024-03-12
System config theme

TJxiaobao:
## What's changed?

The drop-down box is implemented, but there is a problem with rendering
i need help

## Checklist

- [x]  I hereby agree to the terms of the [HertzBeat CLA](https://gist.github.com/tomsun28/511c04e7643901cb550bb6ecc75a661b)
- [ ]  I have read the [Contributing Guide](https://hertzbeat.com/docs/others/contributing/)

## Add or update API

- [ ] I have added the necessary [e2e tests](../e2e) and all cases have passed.


tomsun28:
hi, the new commit [fix save theme prop](https://github.com/dromara/hertzbeat/pull/1636/commits/8270ca9f3ec47e8068bf407a65a78760b5621f9e) has fix prop bing error. 

After that, we need to trigger the actual switching of the page color theme after choosing to switch the theme.

tomsun28:
We can refer the `alain switch theme button` code https://github.com/ng-alain/delon/blob/master/packages/theme/theme-btn/theme-btn.component.ts

TJxiaobao:
Okay, I'll try it

TJxiaobao:
I need help.
I have now implemented the theme switching function, but I encountered a problem.
Because our theme switching function requires passing fields, the page will be refreshed every time when switching to save another theme. It is at this time that refreshing the page causes the theme to change back to the original default theme color. I also thought about and referred to some methods, but failed to solve this problem. Now I need some suggestions or some help. Thanks
如图：
image
image
image
image

tomsun28:
hi，主题信息持久化这个我想我们可以通过 localstorage 和 数据库来做，在之前的颜色主题切换中，主题颜色信息是被存储到了web localstorage的一个字段里面了，关于你切换主题后有马上刷新变回来，我感觉就是之前那个字段导致的，需要我们也去变更下那个。
还有就是我们可以在页面渲染的时候，StartupService 服务里面去主动拉取后端保持的主题配置，然后调用切换主题方法。

hertzbeat:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
Hi, I think we can persist theme information through localstorage and database. In the previous color theme switching, the theme color information was stored in a field of web localstorage. Regarding the change that will be refreshed immediately after you switch the theme. Coming back, I feel that it was caused by the previous field, and we need to change that one too.
Also, when the page is rendered, we can actively pull the theme configuration maintained by the backend in the StartupService service, and then call the switching theme method.

TJxiaobao:
Okay, I'll try it

PR #1683, opened 2024-03-24
support config common aes secret

tomsun28:
## What's changed?

support config common aes secret

## Checklist

- [x]  I hereby agree to the terms of the [HertzBeat CLA](https://gist.github.com/tomsun28/511c04e7643901cb550bb6ecc75a661b)
- [ ]  I have read the [Contributing Guide](https://hertzbeat.com/docs/others/contributing/)

## Add or update API

- [ ] I have added the necessary [e2e tests](../e2e) and all cases have passed.

### DIFF
--- a/alerter/src/main/java/org/dromara/hertzbeat/alert/AlerterProperties.java
+++ b/alerter/src/main/java/org/dromara/hertzbeat/alert/AlerterProperties.java
@@ -40,36 +40,36 @@ public class AlerterProperties {
     /**
      * WeWork webhook url
      */
-    private String weWorkWebHookUrl = "https://qyapi.weixin.qq.com/cgi-bin/webhook/send?key=";
+    private String weWorkWebhookUrl = "https://qyapi.weixin.qq.com/cgi-bin/webhook/send?key=";
 
     /**
      * DingDing talk webhook url
      */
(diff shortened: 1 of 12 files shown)


tomsun28:
todo, consider not init cipher in AesUtil every time and test performance

ZY945:
> todo, consider not init cipher in AesUtil every time and test performance

Does this mean using singleton (not every initialization)?

tomsun28:
> Does this mean using singleton (not every initialization)?

image

here we init `cipher` every time, want to try just initialize the `cipher` once and then support its use by multiple threads safely.

ZY945:
> > Does this mean using singleton (not every initialization)?
> 
> image
> here we init `cipher` every time, want to try just initialize the `cipher` once and then support its use by multiple threads safely.

image
I found a place to set the SecretKey, and we can try initializing Cipher by putting it into the setDefultSecretKey method.

ZY945:
> > I found a place to set the SecretKey, and we can try initializing Cipher by putting it into the setDefultSecretKey method.
> 
> yes, and we should consider the situation that not `setDefultSecretKey` when startup.

image
image
image
Is this method okay? A single instance of double check lock

ZY945 (approved):
LGTM

PR #1787, opened 2024-04-19
[Improve]When multiple lines are returned, you can optionally determine whether only the first line triggers the alarm

15613060203:
When multiple lines are returned, you can optionally determine whether only the first line triggers the alarm;


tomsun28:
👍 Hi, sometimes having too many options to configure is not a good thing for users. They don’t understand what this option does. It is recommended that we can do like other alarm systems, trigger alarms for all rows instead.

15613060203:
> 👍 Hi, sometimes having too many options to configure is not a good thing for users. They don’t understand what this option does. It is recommended that we can do like other alarm systems, trigger alarms for all rows instead.

Agreed. Currently, when multiple lines are returned and the alarm conditions are met, only the first line is triggered and the remaining lines are ignored. If the value is set to trigger all, a monitoring item may trigger multiple alarms. It depends on the user's usage scenario.

tomsun28:
👍 Recommend that we modify it to trigger all rows.

Issue #1625, opened 2024-03-06
[Task] <Support discovery monitoring instance through http_sd>


Calvin979:
### Description

1. http request http_sd api to get service instance ip, port and more.
2. auto create monitor to monitoring service instance throght prometheus protocol

Refer:
https://prometheus.io/docs/prometheus/latest/configuration/configuration/#http_sd_config
https://prometheus.io/docs/prometheus/latest/http_sd/

Calvin979:
@tomsun28 Please assign this issue to me.

Calvin979:
> Hi, I feel that we need to carefully consider the following designs
> 
> * What method is used to call the http_sd api to obtain service instance? Is it a scheduled request or a one-time request? When to update service instance data.
> * How to automatically create monitoring.
> * more

Thanks for your advice. I will check the code and consider how to design it better.

Calvin979:
@tomsun28 Hi, I found that `Prometheus` uses `http_sd` or `file_sd` to discover service instances automatically in order to manage them better. Here is the illustration: 
![image](https://github.com/dromara/hertzbeat/assets/131688897/b9da5169-2889-4b31-98b1-0bc7d0bc68ad)

**Do you mean that Hertzbeat needs to support this feature?**
Like when somebody want to creat a RocketMQ monitor, he/she could enter a http_sd endpoint to manage RocketMQ cluster instead of creating monitor for each RocketMQ instances.
Here is the illustration:
![image](https://github.com/dromara/hertzbeat/assets/131688897/3608be4f-b56b-44a1-a49c-f8d1bdb166d9)

tomsun28:
Hi, it’s like this👍👍. We discover and create monitoring instances automatically throght the http_sd endpoint response content. 

I thinks there is a difficulty that to design how and when to create monitoring instances and how to dispaly them.

Calvin979:
Hi, @tomsun28
I've spended some times for sorting process of collecting data and found that there are some existed design problems.

1. All protocols in path `org.dromara.hertzbeat.common.entity.job.protocol` have no super class or interface, which is hard to define and modify base field or config in runtime.
2. Every single concrete collect class in path `org.dromara.hertzbeat.collector.collect` has a lot of duplicate code **due to** `org.dromara.hertzbeat.collector.collect.AbstractCollect` has only one method to normative behavior of subclass.
   Such as `preCheckParam`, `Remote connecting`, `Exception handling` and so on.

Thus, here are my thoughts:
[...]

PR #1791, opened 2024-04-19
Feature: Support discovery monitoring instance through http_sd

Calvin979:
## What's changed?

Corresponding issue:
#1625 

Service discovery can be available for all collectors now and I provide http_sd as a deafult strategy.  
There's something different between one time task and cyclic task while fetching instance info:
1. one time task
image
2. cyclic task
image

**How to add service discovery for collector?**
1. implements `CommonProtocol` interface
image
2. add param field like http_sd in yml
image
~~~yml
  - field: http_sd
    name:
      zh-CN: http_sd
      en-US: http_sd
    type: text
    placeholder: Enter http_sd url
    required: false
    hide: true
~~~
3. override method `getSupportProtocolClass` and point out which protocol is used for current collector
image

**Test result**
01测试连接成功
02监控详情

**What's more**
the priority of sd is higher than host, which means it will use sd first if you config host and http_sd simultaneously.

## Checklist

## Add or update API


Calvin979:
I haven't add sd for any collectors, it needs to be discussed which collector should support for sd

tomsun28:
👍👍 This design implementation is eye-catching and is a good idea. I don't seem to find the part that automatically discovers and creates multiple instances through the information returned by the http_sd interface. For example, http_sd returns ip1-port1 ip2-port2 ip3-port3. We create three new monitoring instances based on these three automatic discoveries.

Calvin979:
@tomsun28 
I'm not sure whether you mean we need to create monitor task automatically. If so, the process is different.
Monitor task must be created and managed by `Manager`, then collector should collect sd data as a cyclic task like this:
image


What I coded in this pr is **update host and port automatically**.
image
It will be helpful for monitoring services that are supported for automatic failover like `Redis sentinel`.

tomsun28:
hi yes, the first design like this as same as your.


image

image
 

Also i think this pr design is great, provides another way of thinking. how about combine the two. Our initial goal is to support automatic maintenance of batch monitoring tasks through the data returned by http_sd.

Calvin979:
@tomsun28 
Thanks for your design. It is a more clear design than my design. Should I modify code in this pr directly?🤣🤣

tomsun28:
👍👍 okay okay. It is not urgent, depends on your time.  Recommend to release an first apache version first and after then consider integrate this big feature.

Calvin979:
> 👍👍 okay okay. It is not urgent, depends on your time. Recommend to release an first apache version first and after then consider integrate this big feature.

OK, I will close this pr. And I will open another pr after first apache version released and I finished this new feature.

PR #1797, opened 2024-04-20
[Improve]When multiple lines are returned, each alarm is triggered instead of only the first alarm

15613060203:
When multiple lines are returned, each alarm is triggered instead of only the first alarm


tomsun28 (approved):
LGTM！Btw, can you help add some unit test for alarm calculate,  feel this is very important for alarm module.

PR #1827, opened 2024-04-24
Code Simplification, Structure Changes, and Translation Work, Along with a Question

dukbong:
#### Description
This PR simplifies the code by using Assert methods and changes the location of the `validateParams()` process.
By moving this process, `validateParams()` can perform a null check earlier.

#### Question
In the JmxCollectImpl code, the `getConnectSession()` checks whether `jmxProtocol.getUrl() != null`, which is already validated in `validateParams()`.
This redundancy seems unnecessary, but I'm not entirely sure if I'm missing something, so I left the source code unchanged.
Do you think this extra check is required?

#### Other Work
This PR also includes translation work, converting text from Chinese to English.

## Checklist

## Add or update API


tomsun28:
hi this is not redundancy. `jmxProtocol.getUrl()` is user cutom url, can equlas null. 
The `validateParams` process is to check url whether has sepcial word when it not null.

dukbong:
> hi this is not redundancy. `jmxProtocol.getUrl()` is user cutom url, can equlas null. The `validateParams` process is to check url whether has sepcial word when it not null.

Thanks for confirming!
I had a feeling there might be something more to it, but I wasn't sure, so I wanted to ask.

tomsun28 (approved):

PR #1850, opened 2024-04-25
[bugfix] when url is null, "url.startsWith" throws NPE bug.

liyin:


collector/src/main/java/org/apache/hertzbeat/collector/collect/nginx/NginxCollectImpl.java
```diff
@@ -139,7 +139,7 @@ private void validateParams(Metrics metrics) throws Exception {
         
         String url = nginxProtocol.getUrl();
         
-        if (StringUtils.isEmpty(url) || !url.startsWith(RIGHT_DASH)) {
+        if (url == null || StringUtils.isEmpty(url) || !url.startsWith(RIGHT_DASH)) {
```
crossoverJie, line 142: image

The method `StringUtils.isEmpty(url)` has already checks whether string is null or empty.

So maybe we don't need this check `url == null`.

collector/src/main/java/org/apache/hertzbeat/collector/collect/http/SslCertificateCollectImpl.java
```diff
@@ -159,7 +159,7 @@ private void validateParams(Metrics metrics) throws Exception {
         }
         HttpProtocol httpProtocol = metrics.getHttp();
         String url = httpProtocol.getUrl();
-        if (StringUtils.hasText(url) || !url.startsWith(RIGHT_DASH)) {
+        if (url == null || StringUtils.hasText(url) || !url.startsWith(RIGHT_DASH)) {
```
crossoverJie, line 162: I suggest that the same function:`StringUtils.isEmpty(url)` can be used for both of these string checks.

crossoverJie:
Could you please add some unit tests for this?

You can refer this:
https://github.com/apache/hertzbeat/blob/8ffe0dae5c6839ab46cfd3449e643c4101ae614f/collector/src/test/java/org/apache/hertzbeat/collector/collect/nginx/NginxCollectImplTest.java#L75

---
BTW, it is recommended to fill in the content according to the PR template. If the content is empty, it will not be conducive to code review.

PR #1853, opened 2024-04-26
[doc] translates chinese comment to english.

dukbong:
## What's changed?

This pull request translates Chinese to English.

### Question.
It looks like many methods are recording the start time and measuring the response time. Could I get your thoughts on using AOP to store the start time in ThreadLocal when entering the method, then using that to handle response time calculations? I'm still learning, so I'd appreciate your advice.

If it's okay, I'd like to give it a try.

## Checklist

## Add or update API


tomsun28:
hi, can you tell me which method? AOP may not be suitable in some places.

tomsun28 (approved):
LGTM

PR #1801, opened 2024-04-21
[improve] use eclipselink orm replace of hibernate orm

tomsun28:
## What's changed?

use eclipselink orm replace of hibernate orm, due hibernate core licenses is lgpl. 

eclipselink is another implement for jpa guide

## Checklist

- [x]  I have read the [Contributing Guide](https://hertzbeat.apache.org/docs/community/code_style_and_quality_guide)
- [ ]  I have written the necessary doc or comment.
- [ ]  I have added the necessary unit tests and all cases have passed.

## Add or update API

- [ ] I have added the necessary [e2e tests](https://github.com/apache/hertzbeat/tree/master/e2e) and all cases have passed.

PR #1854, opened 2024-04-26
[Improve]Fixed a template modification exception when no monitoring items are running in the monitoring template

15613060203:
Fixed a template modification exception when no monitoring items are running in the monitoring template


tomsun28:
hi, I have not reproduced it locally. I guess it is because use eclipselink orm replace of hibernate orm https://github.com/apache/hertzbeat/pull/1801 . This requires certain table structure changes and data changes. Such problems will occur when using old databases. I will provide an automatic upgrade script later.  you can try using an empty database instead. 

image

PR #1973, opened 2024-05-13
[feature]Hertzbeat custom plugin.

zqr10159:
## What's changed?

Currently Hertzbeat in use , mainly rely on the alert module to notify the user , and then the user to take a number of measures such as sending a request , the execution of sql , the execution of shell scripts , etc. , but can only be manually or webhook to receive the alert information for automated processing , this pr add a plugin module , the module has an interface Plugin, the user can implement this interface execute method to receive Alert as a parameter for customized operations. After writing the code, users only need to package the plugin module, copy it to the /ext-lib folder in the installation directory, restart the main hertzbeat program, you can realize the custom function of executing the alert, without having to re-package and deploy the whole program.

## Checklist

## Add or update API


plugin/src/main/java/org/apache/hertzbeat/plugin/Plugin.java
```diff
@@ -0,0 +1,28 @@
...
+
+
+package org.apache.hertzbeat.plugin;
+
+import org.apache.hertzbeat.common.entity.alerter.Alert;
+
+/**
+ * Plugin
+ */
+public interface Plugin {
```
crossoverJie, line 26: Is this interface just for custom alerting? If so, it is recommended that the name be changed to `AlertPlugin`.

manager/src/main/java/org/apache/hertzbeat/manager/component/alerter/DispatcherAlarm.java
```diff
@@ -113,6 +118,13 @@ public void run() {
                         alertStoreHandler.store(alert);
                         // Notice distribution
                         sendNotify(alert);
+                        // Execute the plugin
+                        List<Object> beans = pluginConfig.getBean();
+                        for (Object bean : beans) {
+                            if (bean instanceof Plugin) {
```
crossoverJie, line 124: Is it possible to use generics here to avoid type conversion?

manager/src/main/java/org/apache/hertzbeat/manager/config/PluginConfig.java
```diff
@@ -0,0 +1,177 @@
...
+        if (files == null) {
+            return null;
+        }
+        return Arrays.stream(files)
+                .filter(f -> f.getName().endsWith(".jar"))
+                .map(File::getAbsolutePath)
+                .collect(Collectors.toList());
+    }
+
+    private List<String> getAllClassNamesFromJar() throws IOException {
```
crossoverJie, line 102: Is it possible to use `SPI` to simplify `classloader`.

manager/src/main/java/org/apache/hertzbeat/manager/config/PluginConfig.java
```diff
@@ -0,0 +1,177 @@
...
+                        registerBean(clazz, registry);
+                    }
+                }
+            }
+        }    }
+
+    public List<Object> getBean() throws IOException {
+        List<String> names = getAllClassNamesFromJar();
+        List<Object> beans = new ArrayList<>();
+        if (names != null) {
```
crossoverJie, line 162: The enhanced for loop may no longer need to check whether `name != null`.

crossoverJie (approved):
Gread job, I left a few comments so we can discuss it.

plugin/src/main/java/org/apache/hertzbeat/plugin/Plugin.java
zqr10159: Just a preliminary implementation, the current timing is only after the alarm, the follow-up can be added to an enumeration class to specify the timing, such as the collection, triggering the alarm silence, triggering the alarm convergence, etc.

manager/src/main/java/org/apache/hertzbeat/manager/component/alerter/DispatcherAlarm.java
zqr10159: Please help me modify it.

manager/src/main/java/org/apache/hertzbeat/manager/config/PluginConfig.java
zqr10159: Please help me modify it.

manager/src/main/java/org/apache/hertzbeat/manager/config/PluginConfig.java
zqr10159: Please help me modify it.

plugin/src/main/java/org/apache/hertzbeat/plugin/Plugin.java
crossoverJie: What I mean is that it is best to define the interface at the beginning, otherwise subsequent changes may involve break changes, leading to compatibility issues.

```java
public interface Plugin {
    void execute(Alert alert);
}
```

The interface's name is `Plugin`, and the method is `execute(Alert alert)`.

They are all common names but they only do one thing: alerting. Maybe they can be more focused, such as:

```java
public interface AlertPlugin {
     void execute(Alert alert);
}

// or
public interface Plugin {
     void alert(Alert alert);
}
```

manager/src/main/java/org/apache/hertzbeat/manager/config/PluginConfig.java
crossoverJie: This is my mistake, but it is still recommended to use `Optional` to reduce some conditional judgments.

```java
Optional.ofNullable(beans). ifPresent(()->{
// dosomthing
})

// or retrun Optional
Optional<List<String>> names = getAllClassNamesFromJar()
```

tomsun28 (approved):
👍LGTM

crossoverJie:
Gread job, LGTM.

Some documentation will need to be added later.

zqr10159:
> Gread job, LGTM.
> 
> 
> 
> Some documentation will need to be added later.

I will

PR #1991, opened 2024-05-14
[improve] optimize collector httpsd discovery

yuluo-yx:
- Optimize collector httpsd discovery


manager/src/main/resources/define/app-http_sd.yml
```diff
@@ -47,7 +47,7 @@ params:
     # type-param field type(most mapping the html input type)
     type: number
     # when type is number, range is required
-    range: '[0,65535]'
+    range: "[0,65535]"
```
tomsun28, line 50: hi, here we use 'xxx'

manager/src/main/resources/define/app-http_sd.yml
```diff
@@ -92,7 +92,7 @@ metrics:
     protocol: httpsd
     httpsd:
       host: ^_^host^_^
-      port: ^_^port^_^
+      port: 1234
```
tomsun28, line 95: hi, why pined the port number 1234

common/src/main/java/org/apache/hertzbeat/common/entity/job/protocol/HttpsdProtocol.java
```diff
@@ -34,10 +34,10 @@
 @NoArgsConstructor
 public class HttpsdProtocol {
     private String host;
-    private String port;
+    private int port;
```
tomsun28, line 37: the port type in yml need string, due we need string param like `^_^port^_^`

manager/src/main/resources/define/app-http_sd.yml
yuluo-yx: done!

manager/src/main/resources/define/app-http_sd.yml
yuluo-yx: this is the random number. only example.

common/src/main/java/org/apache/hertzbeat/common/entity/job/protocol/HttpsdProtocol.java
yuluo-yx: In `HttpsdProtocol.java`, I've changed `string` to `int` to be consistent with the port database type in the service discovery component.

tomsun28: yes, but the param `port` is custom, user can custom the port value with `^_^port^_^`, if here we input the `^_^port^_^` and port type is int, will casue error.

manager/src/main/resources/define/app-http_sd.yml
tomsun28: can not use a example number instead of ^_^port^_^ param

common/src/main/java/org/apache/hertzbeat/common/entity/job/protocol/HttpsdProtocol.java
yuluo-yx: got it, I will restore it.

tomsun28:
hi @Calvin979 how about this optimize httpsd pr?

Calvin979:
> hi @Calvin979 how about this optimize httpsd pr?

It is better than before!

tomsun28:
> @tomsun28 if this pr has any problem, pls tell me.

Sorry for the delay, I will review tonight

tomsun28 (approved):
LGTM!

PR #2021, opened 2024-05-20
[fix] Fix possible potential thread safe bugs

gjjjj0101:
## What's changed?
Refactored some code to ensure thread safety. And I think `addNode()` and `removeNode()` in `ConsistentHash` class may cause thread unsafe issues.  Please give some suggestions.

## Checklist

## Add or update API

### DIFF
--- a/collector/src/main/java/org/apache/hertzbeat/collector/collect/common/cache/ConnectionCommonCache.java
+++ b/collector/src/main/java/org/apache/hertzbeat/collector/collect/common/cache/ConnectionCommonCache.java
@@ -197,15 +197,18 @@ public class ConnectionCommonCache<T, C extends AbstractConnection<?>> {
             cacheMap.remove(key);
             return Optional.empty();
         }
-        C value = cacheMap.get(key);
-        if (value == null) {
-            log.error("[connection common cache] value is null, remove it, key {}.", key);
-            cacheMap.remove(key);
-            timeoutMap.remove(key);
-        } else if (refreshCache) {
-            cacheTime[0] = System.currentTimeMillis();
-            timeoutMap.put(key, cacheTime);
-        }
+        C value = cacheMap.compute(key, (k, v) -> {
+            if (v == null) {
+                log.error("[connection common cache] value is null, remove it, key {}.", key);
+                timeoutMap.remove(key);
+                return null;
+            }
+            if (refreshCache) {
+                cacheTime[0] = System.currentTimeMillis();
+                timeoutMap.put(key, cacheTime);
+            }
+            return v;
+        });
         return Optional.ofNullable(value);
     }
 
--- a/collector/src/test/java/org/apache/hertzbeat/collector/collect/common/cache/CommonCacheTest.java
+++ b/collector/src/test/java/org/apache/hertzbeat/collector/collect/common/cache/CommonCacheTest.java
@@ -17,40 +17,91 @@
 
 package org.apache.hertzbeat.collector.collect.common.cache;
 
-import org.junit.jupiter.api.AfterEach;
+import java.util.Optional;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import static org.junit.jupiter.api.Assertions.*;
 
 /**
  * Test case for {@link ConnectionCommonCache}
  */
+@ExtendWith(MockitoExtension.class)
 class CommonCacheTest {
+
+    @Mock
+    private AbstractConnection<?> mockConnection;
+
+    @InjectMocks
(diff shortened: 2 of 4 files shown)


collector/src/main/java/org/apache/hertzbeat/collector/collect/common/cache/ConnectionCommonCache.java
```diff
@@ -197,15 +197,18 @@ public Optional<C> getCache(T key, boolean refreshCache) {
...
-        C value = cacheMap.get(key);
-        if (value == null) {
-            log.error("[connection common cache] value is null, remove it, key {}.", key);
-            cacheMap.remove(key);
-            timeoutMap.remove(key);
-        } else if (refreshCache) {
-            cacheTime[0] = System.currentTimeMillis();
-            timeoutMap.put(key, cacheTime);
-        }
+        C value = cacheMap.compute(key, (k, v) -> {
```
crossoverJie, line 200: LGTM, please add a unit test.

crossoverJie (commented):
Great job, please add unit tests for these changes.

TJxiaobao:
I think it is possible to use a global lock to solve this problem, the amount of concurrent addnode removenode will not greatly affect performance.

collector/src/main/java/org/apache/hertzbeat/collector/collect/common/cache/ConnectionCommonCache.java
gjjjj0101: Thanks, I have added unit tests. Please review.

gjjjj0101:
> I think it is possible to use a global lock to solve this problem, the amount of concurrent addnode removenode will not greatly affect performance.

How about divides the existing `addNode` and `removeNode` functions into smaller, synchronized operations: `addVMNode` and `removeVMNode`? In this way,  `addNode` and `removeNode` also can be called in parallel.

collector/src/test/java/org/apache/hertzbeat/collector/collect/common/cache/CommonCacheTest.java
```diff
@@ -17,40 +17,91 @@
...
+import static org.junit.jupiter.api.Assertions.*;
 
 /**
  * Test case for {@link ConnectionCommonCache}
  */
+@ExtendWith(MockitoExtension.class)
 class CommonCacheTest {
 
+    @Mock
+    private AbstractConnection<?> mockConnection;
```
crossoverJie, line 36: `ConnectionCommonCache` does not depend on external services, so there is no need to use `mock`.

Here you can create an anonymous `AbstractConnection` class for testing.

collector/src/main/java/org/apache/hertzbeat/collector/collect/common/cache/ConnectionCommonCache.java
```diff
@@ -197,15 +197,18 @@ public Optional<C> getCache(T key, boolean refreshCache) {
             cacheMap.remove(key);
             return Optional.empty();
         }
-        C value = cacheMap.get(key);
-        if (value == null) {
-            log.error("[connection common cache] value is null, remove it, key {}.", key);
-            cacheMap.remove(key);
```
crossoverJie, line 203:
```java
cacheMap.remove(key);
```

Why was this line of code deleted?

gjjjj0101: Because `cacheMap` uses `ConcurrentLinkedHashMap<K, V>`. The implementation of `put()` method in `ConcurrentLinkedHashMap<K, V>` checks for null value and throws `NullPointerException` if value is null. So `cache.get(key) == null` is true only when the key does not exist in the map.

collector/src/test/java/org/apache/hertzbeat/collector/collect/common/cache/CommonCacheTest.java
gjjjj0101: Thanks, PTAL.

TJxiaobao:
> > I think it is possible to use a global lock to solve this problem, the amount of concurrent addnode removenode will not greatly affect performance.
> 
> How about divides the existing `addNode` and `removeNode` functions into smaller, synchronized operations: `addVMNode` and `removeVMNode`? In this way, `addNode` and `removeNode` also can be called in parallel.

Sorry to reply today, personally think this idea is also very good, you can try it.

tomsun28 (approved):
👍👍LGTM！

collector/src/main/java/org/apache/hertzbeat/collector/collect/common/cache/ConnectionCommonCache.java
ben-manes: Please consider upgrading from CLHM to Caffeine at some opportune time. Thanks!

gjjjj0101: > Please consider upgrading from CLHM to Caffeine at some opportune time. Thanks!

Thanks, we will upgrade to Caffeine in the near future！

PR #2044, opened 2024-05-28
[improve] Add test for MemcachedCollectImpl

zuobiao-zhou:
## What's changed?

Add test for MemcachedCollectImpl.

## Checklist


collector/src/test/java/org/apache/hertzbeat/collector/collect/memcached/MemcachedCollectImplTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Test case for {@link MemcachedCollectImpl}
+ */
+@ExtendWith(MockitoExtension.class)
+public class MemcachedCollectImplTest {
+    @Mock
+    MemcachedProtocol memcachedProtocol;
+
+    @Mock
```
gjjjj0101, line 50: hi, I think it's no need to mock `metrics` and `builder`. Because it just create a real object passed as a parameter to the method, the method uses the real object and does not need to simulate the object behavior. Just use it as a variable☺️.

tomsun28 (approved):
lgtm!

PR #2067, opened 2024-06-05
[collector] fix the wrong condition for checking the response

ileonli:
## What's changed?

We should log the information only when the `resp` does not include text.

It appears that the condition used to check the response is incorrect.

## Checklist


collector/src/main/java/org/apache/hertzbeat/collector/collect/http/HttpCollectImpl.java
```diff
@@ -133,7 +133,7 @@ public void collect(CollectRep.MetricsData.Builder builder,
             // Option 1: Parse using InputStream, but this requires significant code changes;
             // Option 2: Manually trigger garbage collection, similar to how it's done in Dubbo for large inputs.
             String resp = EntityUtils.toString(response.getEntity(), StandardCharsets.UTF_8);
-            if (StringUtils.hasText(resp)) {
+            if (!StringUtils.hasText(resp)) {
                 log.info("http response entity is empty, status: {}.", statusCode);
```
crossoverJie, line 137: When `resp` is empty, should we return directly without executing the subsequent logic?

tomsun28: some http request may not has response data themselves, but still need to be processed.

tomsun28 (approved):

PR #2120, opened 2024-06-22
[feature] support ai api with zhipu ai 

Yanshuming1:
add ai feature


zqr10159:
Can you introduce it in detail?

Yanshuming1:
> Can you introduce it in detail?

By configuring AI config in application.yml.yml, users can directly access ZhipuAI through http, and the back end streams back with sse

Yanshuming1:
I will supplement the document later

manager/pom.xml
```diff
@@ -106,6 +106,13 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-thymeleaf</artifactId>
         </dependency>
+
+        <!--webflux-->
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-webflux</artifactId>
+        </dependency>
+
```
tomsun28, line 115: hi, due we add new thrid denpedency, we need update the license doc here. https://github.com/apache/hertzbeat/tree/master/material/licenses

tomsun28 (approved):
👍 lgtm thanks.

manager/src/main/java/org/apache/hertzbeat/manager/service/impl/ZhiPuServiceImpl.java
```diff
@@ -0,0 +1,118 @@
...
+                .defaultHeader(HttpHeaders.AUTHORIZATION, "Bearer " + apiKey)
+                .exchangeStrategies(ExchangeStrategies.builder()
+                        .codecs(item -> item.defaultCodecs().maxInMemorySize(16 * 1024 * 1024))
+                        .build())
+                .build();
+    }
+
+    @Override
+    public String getType() {
+        return "0";
```
tomsun28, line 70: hi, suggest use string  or enum `zhipu` instead of number `0`

Yanshuming1: It already is.

PR #2160, opened 2024-06-29
Fixed notify module error

kerwin612:
## What's changed?

before:
bug

after:
fixed

Originally, the notification center referenced the `notice-icon` component, which was implemented based on the `nz-tabs` component. However, there was a bug where the **tab-panel** would not show up in **hertzbeat** when there was only one tab. Given the current usage scenario, we can avoid relying on the ·notice-icon· component and implement it ourselves.

## Checklist

## Add or update API


tomsun28 (approved):
thanks lgtm.

PR #2180, opened 2024-07-03
[Improve] Refactor the notification sending logic.

crossoverJie:
Refactor the notification sending logic.

I noticed that most of the `AlertNotifyHandlers` require the use of `NoticeTemplate`, so I directly filtered out the data with null `templateId`, making the code more concise.

## Checklist

## Add or update API


yuluo-yx (approved):
lgtm

manager/src/main/java/org/apache/hertzbeat/manager/component/alerter/DispatcherAlarm.java
```diff
@@ -131,29 +132,21 @@ public void run() {
...
-                        }
-                    } catch (AlertNoticeException e) {
-                        log.warn("DispatchTask sendNoticeMsg error, message: {}", e.getMessage());
-                    }
-                }
-            }
+            matchNoticeRulesByAlert(alert).ifPresent(noticeRules -> {
+                noticeRules.forEach(rule -> {
+                    workerPool.executeNotify(() -> {
+                        rule.getReceiverId().stream().filter(receiverId -> rule.getTemplateId() != null)
```
tomsun28, line 138: hi, the code logic here is to filter out cases where the template id is null, but before the template id was null it would go to use the default file notification template.

crossoverJie: Thank you for the reminder. I adjusted the logic; 
I moved the template query earlier so it can return exceptions faster, and the code can be more concise.

tomsun28 (approved):
lgtm👍

PR #2251, opened 2024-07-09
[fixbug]fix parameter parsing problem

### LINKED ISSUE #2250: [BUG] Failed to add template

### Current Behavior

Failed to add template, caused by PR # 2122
image

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
HertzBeat version(s):master
```

### Debug logs

_No response_

### Anything else?

_No response_

zhangshenghang:
## What's changed?

fix #2250 

Modify Jackson to ignore the position field instead of throwing an exception directly to make the system unusable

## Checklist


tomsun28:
hi, what about this prop, should we conifig it in jackson ObjectMapper

```
  jackson:
    default-property-inclusion: ALWAYS
```

image

zhangshenghang:
> hi, what about this prop, should we conifig it in jackson ObjectMapper
> 
> ```
>   jackson:
>     default-property-inclusion: ALWAYS
> ```
> 
> image

I will modify it

zhangshenghang:
@tomsun28 

## Modifications
Customized ObjectMapper Beans may override Spring Boot's default Jackson configuration.
Use the Jackson2ObjectMapperBuilder provided by Spring to create the ObjectMapper.

## Verification
When set to
```
jackson:
default-property-inclusion: ALWAYS
deserialization:
fail-on-unknown-properties: true
```
Error
image

---

When configured to
```
jackson:
default-property-inclusion: ALWAYS
deserialization:
fail-on-unknown-properties: false
```
Run successfully.

tomsun28:
👍 /cc @Calvin979

zqr10159:
Can application.yml be left unmodified? Unless it's a user-definable item, it's recommended that it be implemented in the code so that users can avoid some problems when upgrading

tomsun28:
> Can application.yml be left unmodified? Unless it's a user-definable item, it's recommended that it be implemented in the code so that users can avoid some problems when upgrading

hi, yes here the old version upgrade no need modify the application.yml due the code has config too. Is that true? @zhangshenghang

zhangshenghang:
> hi, yes here the old version upgrade no need modify the application.yml due the code has config too. Is that true?



> > Can application.yml be left unmodified? Unless it's a user-definable item, it's recommended that it be implemented in the code so that users can avoid some problems when upgrading
> 
> hi, yes here the old version upgrade no need modify the application.yml due the code has config too. Is that true? @zhangshenghang

The code modified in this PR is implemented using 'Jackson2OrderMapperBuilderCustomizer', and can run normally without setting this value in the configuration file. Not setting the value of 'fail on unknown properties' has the same effect as setting the value of' fail on unknown properties: false '. Only when set to true will the above error be prompted.
```
deserialization:
fail-on-unknown-properties: false
```
﻿
If we want to be compatible with older versions, we can remove 'fail-on-unknown-properties: false'
﻿

tomsun28 (dismissed):

zhangshenghang:
> > hi, yes here the old version upgrade no need modify the application.yml due the code has config too. Is that true?
> 
> > > Can application.yml be left unmodified? Unless it's a user-definable item, it's recommended that it be implemented in the code so that users can avoid some problems when upgrading
> > 
> > 
> > hi, yes here the old version upgrade no need modify the application.yml due the code has config too. Is that true? @zhangshenghang
> 
> The code modified in this PR is implemented using 'Jackson2OrderMapperBuilderCustomizer', and can run normally without setting this value in the configuration file. Not setting the value of 'fail on unknown properties' has the same effect as setting the value of' fail on unknown properties: false '. Only when set to true will the above error be prompted.
> 
> ```
> deserialization:
> fail-on-unknown-properties: false
> ```
> 
> ﻿ If we want to be compatible with older versions, we can remove 'fail-on-unknown-properties: false' @tomsun28

@tomsun28 
Do we need to delete this？
```
deserialization:
 fail-on-unknown-properties: false
```

tomsun28:
> Do we need to delete this？

ok, 👍

zqr10159 (approved):

Issue #2188, opened 2024-07-04
[Task] Customizable Refresh Intervals for Monitoring Metrics


yangshuDBA:
### Description

The current system only allows setting a unified refresh interval for monitoring assets such as the processor, memory, and hard disk of the host. However, in practice, these metrics often require different refresh intervals; for example, disk monitoring may need to refresh every 60 seconds, while processor monitoring may need to refresh every 10 seconds.

To address this, we propose enhancing the monitoring template to allow custom refresh intervals for individual metrics. If a specific interval is not defined for a metric, it should default to the standard refresh interval set for the asset.
![image](https://github.com/apache/hertzbeat/assets/112070510/695dfdd8-567f-4cca-a47a-a0fa7fd32eca)

Monitor template
![image](https://github.com/apache/hertzbeat/assets/112070510/440541df-629d-44e3-be50-1d875112f8c8)

### Task 2: Support Default Refresh Intervals for Metrics in Monitoring Templates When Not Defined

zuobiao-zhou:
please assign to me.

PR #2252, opened 2024-07-09
[feature] Support custom refresh intervals for each group of metrics

zuobiao-zhou:
## What's changed?

close https://github.com/apache/hertzbeat/issues/2188

**Support custom refresh intervals for each group of metrics. The specific implementation is as follows**:

Within a `monitor`, `metrics` with `different refresh intervals` are divided into `different jobs`. If a group's metrics do not have a custom refresh interval set, the default refresh interval specified during the monitor's creation will be used.

Implementation result:
image
image
image

## Checklist


tomsun28:
hi, thanks for contributions. 👍

The one monitor mapping one collect job -> many collect jobs is a big change, we need time to think and design carefully. 
This will cause some collection logics on the collector side to fail, such as supporting sequential collection and concurrent collection, using the collection response as a parameter to enter the next collection metrics, etc. System upgrade is also a big hassle. 

Previously, I thought the modification direction was in the collector scheduling part, splitting the scheduling time of multiple metrics tasks in a job corresponding to a monitoring. 

These two directions need to be carefully considered.

zuobiao-zhou:
> hi, thanks for contributions. 👍
> 
> The one monitor mapping one collect job -> many collect jobs is a big change, we need time to think and design carefully. This will cause some collection logics on the collector side to fail, such as supporting sequential collection and concurrent collection, using the collection response as a parameter to enter the next collection metrics, etc. System upgrade is also a big hassle.
> 
> Previously, I thought the modification direction was in the collector scheduling part, splitting the scheduling time of multiple metrics tasks in a job corresponding to a monitoring.
> 
> These two directions need to be carefully considered.

That makes sense, I'll try to implement the second approach.

tomsun28:
hi, i draw a job task relation for refer. The main difficulty here is the scheduling of multiple metrics collection under one job. At present, I have not thought of a good way to support custom scheduling time.

image

PR #2256, opened 2024-07-09
[improve] : Modified loading application.yml

Yanshuming1:
[improve] : Modified loading application.yml


manager/src/main/java/org/apache/hertzbeat/manager/pojo/dto/AiProperties.java
```diff
@@ -0,0 +1,54 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * AiProperties
+ */
+@Configuration
+@ConfigurationProperties(prefix = "ai")
+@Data
+public class AiProperties {
```
tomsun28, line 30: the `AiProperties` is not a dto pojo, it should in config

Yanshuming1: Ok, I'll modify it later

tomsun28 (approved):

PR #2271, opened 2024-07-11
[feature] Support custom refresh intervals for each group of metrics

zuobiao-zhou:
## What's changed?

close https://github.com/apache/hertzbeat/issues/2188

**Support custom refresh intervals for each group of metrics. The specific implementation is as follows**:

1. Construct `a cycle of job refresh interval list` based on the refresh intervals of each metric.
2. When the job dispatches, select metrics that are `due for refresh` to form `priorMetrics`.

Implementation result:
image
image
image

## Checklist


tomsun28:
hi sorry for slow response.  The idea of ​​implementing is great 👍👍

There is a strange bug here when I test.  

I set intervals 30 or 60s when adds `port` monitoring, the collect interval time work success. 

iShot_2024-07-17_09 21 00

iShot_2024-07-17_09 21 43

iShot_2024-07-17_09 21 18

But when i add `ubuntu` monitoring, config the intervals 60s, the real collect interval time is 2*60s. Seem it is caused by multiple metrics collection and scheduling under `ubuntu`, and `port` only has one metric.

iShot_2024-07-17_09 26 45

iShot_2024-07-17_09 26 55

iShot_2024-07-17_09 27 07

zuobiao-zhou:
Hi @tomsun28, thanks for your feedback. I will investigate the cause and fix this strange bug.

PR #2316, opened 2024-07-19
[chore] optimize e2e folder struct

yuluo-yx:
optimize e2e folder struct


tomsun28:
hi personally, I think testsuite.yaml should not be placed in the data directory, just keep the status quo.

tomsun28:
image
hi, is this the reason why move the testsuite.yaml ?

yuluo-yx:
> hi personally, I think testsuite.yaml should not be placed in the data directory, just keep the status quo.

got it. I will restore it.

yuluo-yx:
> image hi, is this the reason why move the testsuite.yaml ?

no, here are some of my personal thoughts, I think these are data configurations and the e2e tests are occasional errors. It can't be reproduced locally, and I'm looking into keeping a running log of the system in github ci.

yuluo-yx:
@tomsun28 restore it, ptal

tomsun28 (approved):

PR #2317, opened 2024-07-20
[Improve] add kafak queue unit test

yuluo-yx:
- Add kafak queue unit test


common/src/main/java/org/apache/hertzbeat/common/entity/message/CollectRep.java
```diff
@@ -449,7 +449,7 @@ public static final class MetricsData extends
         private MetricsData(com.google.protobuf.GeneratedMessageV3.Builder<?> builder) {
             super(builder);
         }
-        private MetricsData() {
+        public MetricsData() {
```
tomsun28, line 452: hi, the `CollectRep` is auto geneate by protocol buffer, can not be edit.

yuluo-yx: sorry, my bad. I will fix it.

yuluo-yx:
@tomsun28 fix it. ptal

tomsun28 (approved):

PR #2333, opened 2024-07-22
[Improve] add connection check when get connection

yuluo-yx:


tomsun28:
hi, the connection cache is the underlying code. If it throws an exception, more upper-level code logic should be considered.

collector/src/main/java/org/apache/hertzbeat/collector/collect/common/cache/JdbcConnect.java
```diff
@@ -39,8 +40,23 @@ public void closeConnection() throws Exception {
...
+    }
+
     @Override
     public Connection getConnection() {
+
+        try {
+            this.check();
+        }
+        catch (SQLException e) {
+            throw new RuntimeException(e.getMessage());
```
yuluo-yx, line 58: Returning null here and printing the logs would probably block the @tomsun28

PR #2371, opened 2024-07-24
[Improve] add database file server provider

pwallk:
## What's changed?

- add local database file server provider
- improve define yml delete

## Checklist

## Add or update API


common/src/main/java/org/apache/hertzbeat/common/entity/manager/Define.java
```diff
@@ -0,0 +1,73 @@
...
+@NoArgsConstructor
+@Schema(description = "monitor define entity")
+@EntityListeners(AuditingEntityListener.class)
+public class Define {
+
+    @Id
+    @Schema(title = "app", example = "websocket")
+    private String app;
+
+    @Column(columnDefinition = "text")
```
tomsun28, line 54: hi,  `columnDefinition = "text"` works with MySQL and PostgreSQL, but less portable. suggest use `@Lob`

common/src/main/java/org/apache/hertzbeat/common/entity/manager/Define.java
```diff
@@ -0,0 +1,75 @@
...
+@Schema(description = "monitor define entity")
+@EntityListeners(AuditingEntityListener.class)
+public class Define {
+
+    @Id
+    @Schema(title = "app", example = "websocket")
+    private String app;
+
+    @Lob
+    @Column(columnDefinition = "text")
```
tomsun28, line 56: @Lob and `@Column(columnDefinition = "text")` are duplicates

common/src/main/java/org/apache/hertzbeat/common/entity/manager/Define.java
pwallk: I'm going to remove that

common/src/main/java/org/apache/hertzbeat/common/entity/manager/Define.java
pwallk: I'm going to remove that

tomsun28 (approved):
lgtm

PR #2372, opened 2024-07-24
[fix] fix committer name

Yanshuming1:
fix committer name


manager/pom.xml
```diff
@@ -145,7 +145,6 @@
         <dependency>
             <groupId>mysql</groupId>
             <artifactId>mysql-connector-java</artifactId>
-            <scope>provided</scope>
         </dependency>
```
tomsun28, line 148: hi, can not depend the mysql jar when package

Yanshuming1: Changed it. Accidentally uploaded the local code

tomsun28 (approved):

PR #2375, opened 2024-07-25
[Improve] add AlertDefineExcel unit test and fix bugs

yuluo-yx:
add AlertDefineExcel unit test and fix bugs


alerter/pom.xml
```diff
@@ -82,6 +82,12 @@
             <version>4.1.1</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
```
yuluo-yx, line 85: @tomsun28 if not add this dependency, writeOS method will not work. so I add it. any ideas?

alerter/src/main/java/org/apache/hertzbeat/alert/service/impl/AlertDefineAbstractImExportServiceImpl.java
```diff
@@ -105,47 +100,4 @@ protected String fileNamePrefix() {
         return "hertzbeat_alertDefine_" + LocalDate.now();
     }
 
-    /**
```
yuluo-yx, line 108: I will move to `dto` pkg.

alerter/pom.xml
tomsun28: hi, the poi-oomxl has been added in licenses doc, you can add it here. 

image

for alarm define excel export, maybe can refer the monitors execl export import

tomsun28 (approved):
👍

PR #2376, opened 2024-07-25
[improve] add some code style improve

Aias00:
## What's changed?

add some code style improve

## Checklist

## Add or update API


collector/src/main/java/org/apache/hertzbeat/collector/collect/common/ssh/CommonSshClient.java
```diff
@@ -43,7 +43,7 @@ public class CommonSshClient {
         PropertyResolverUtils.updateProperty(
                 SSH_CLIENT, CoreModuleProperties.HEARTBEAT_INTERVAL.getName(), 2000);
         PropertyResolverUtils.updateProperty(
-                SSH_CLIENT, CoreModuleProperties.HEARTBEAT_REPLY_WAIT.getName(), 300_000);
+                SSH_CLIENT, CoreModuleProperties.HEARTBEAT_NO_REPLY_MAX.getName(), 300_000);
```
tomsun28, line 46: hi the HEARTBEAT_REPLY_WAIT HEARTBEAT_NO_REPLY_MAX seem value is different. 
it seems HEARTBEAT_REPLY_WAIT is time, HEARTBEAT_NO_REPLY_MAX is times. 

    /**
     * Key used to indicate that the heartbeat request is also expecting a reply - time in <U>milliseconds</U> to wait
     * for the reply. If non-positive then no reply is expected (nor requested).
     *
     * @deprecated since 2.13.0, use {@link #HEARTBEAT_NO_REPLY_MAX} instead
     */
    @Deprecated
    public static final Property<Duration> HEARTBEAT_REPLY_WAIT
            = Property.durationSec("heartbeat-reply-wait", Duration.ofMinutes(5));

    /**
     * Key to set the maximum number of heartbeat messages to send without having received a reply. If &gt; 0, heartbeat
     * messages are sent with a flag that requires the peer to reply. The session will be killed if
     * {@code HEARTBEAT_NO_REPLY_MAX} heartbeats have been sent without having received a reply. If &lt;= 0, heartbeat
     * messages will be sent, but no reply is requested or expected, and the client will not kill the session.
     *
     * @since 2.13.0
     */
    public static final Property<Integer> HEARTBEAT_NO_REPLY_MAX = Property.integer("heartbeat-no-reply-max", 0);

Aias00: sorry i made a mistake, i have changed this property.

tomsun28: hi the note it suggest use HEARTBEAT_NO_REPLY_MAX instead of Deprecated-HEARTBEAT_REPLY_WAIT. How about we set the right HEARTBEAT_NO_REPLY_MAX value for it.

tomsun28: We can open a new PR to do this later, but the HEARTBEAT_NO_REPLY_MAX  value needs to be studied and confirmed.

tomsun28 (approved):
lgtm!

PR #2392, opened 2024-07-27
[Improve] add AlertDefineYamlImExportServiceImpl unit test

yuluo-yx:


alerter/src/test/java/org/apache/hertzbeat/alert/service/AlertDefineYamlImExportServiceTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+		List<ExportAlertDefineDTO> result = service.parseImport(inputStream);
+
+		assertNotNull(result);
+		assertEquals(1, result.size());
+
+		// yaml load data to string result and alertDefineList are not equal. why?
+		// System.out.println(alertDefineList.toString());
+		// System.out.println(result.toString());
+		// assertEquals(alertDefineList, result);
```
Calvin979, line 110: image
Looks like `snakeyaml` use LinkedHashMap as its underlying data structure.
Maybe we can try to compare them in a different way.

yuluo-yx: yep, I try to solve it.

Calvin979 (approved):

PR #2396, opened 2024-07-28
[chore] move dto object to dto package

yuluo-yx:


tomsun28:
hi, It is not recommended to put private dto for a single function into the public dto package.  `org.apache.hertzbeat.manager.component.alerter` contain all messge alerter code, this makes it easier to keep the code cohesive and to extract it later.

yuluo-yx:
got it

Issue #1480, opened 2024-01-10
[Question] Remote ssh server no more session channel, please increase sshd_config MaxSessions


leim:
### Question

centos服务器监控，运行了有一段时间了，最近突然报这个错误，并且监控显示【不可用】，但是通过ssh方式连接服务器是没有问题的。
后来调整hertzbeat的配置【复用连接】，将其关闭，现在又变得好使了。
请问这个【复用连接】的作用是什么。

hertzbeat:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
### Question

Centos server monitoring has been running for a while. Recently, this error suddenly reported, and the monitoring showed [Unavailable], but there is no problem connecting to the server through ssh.
Later, I adjusted the configuration of hertzbeat [Reuse Connection] and turned it off, and now it works fine again.
What is the function of this [reuse connection]?

tomsun28:
【复用连接】就是一直复用保持一个ssh tcp链接，但是长时间运行在某些环境有可能server因为其它原因断开保持 但是client没有察觉 导致出链接失败的现象

hertzbeat:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
[Reuse connection] is to reuse and maintain an ssh tcp link. However, if it runs for a long time in some environments, the server may be disconnected due to other reasons, but the client is not aware of it, resulting in a link failure.

loong95:
We had the same problem in our environment, even after turning off the "Reuse connection" option, we still couldn't connect, and restarting the target server's sshd service didn't work, but only after restarting hertzbeat could we connect!
![image](https://github.com/user-attachments/assets/f8a78c3c-f3e9-48ff-bec4-ae7e24dfea52)

tomsun28:
Got it, we will follow

PR #2398, opened 2024-07-28
[collector]bugfix: Fix ssh session not closing properly

loong95:
## What's changed?

Fix [bug 1480](https://github.com/apache/hertzbeat/issues/1480)

* A cancel signal needs to be sent if the execution times out, otherwise the session cannot be closed promptly.
* Call `ChannelExec.close` with the 'false' parameter to ensure the session is not kept alive.

If you want to reproduce the bug, set an ssh command that takes longer to execute than the timeout duration, such as 'sleep 10'. After the number of timeout occurrences reaches MaxSessions, it will trigger 'no session'.

## Checklist

## Add or update API


tomsun28 (approved):
lgtm!👍

PR #2412, opened 2024-07-30
[Improve] add AlarmCommonReduce unit test

yuluo-yx:
- Add AlarmCommonReduce unit test & fix bugs


alerter/src/main/java/org/apache/hertzbeat/alert/reduce/AlarmCommonReduce.java
```diff
@@ -64,6 +65,7 @@ public void reduceAndSendAlarm(Alert alert) {
                 }
             }
         }
+        System.out.println(tags);
```
tomsun28, line 68: hi not suggest `System.out.println`

yuluo-yx: my bad, this is test output.

tomsun28 (approved):

PR #2426, opened 2024-07-31
[Improve] add alarm define slience unit test

yuluo-yx:


alerter/src/main/java/org/apache/hertzbeat/alert/controller/AlertDefinesController.java
```diff
@@ -111,8 +112,11 @@ public ResponseEntity<Message<Page<AlertDefine>>> getAlertDefines(
         };
         Sort sortExp = Sort.by(new Sort.Order(Sort.Direction.fromString(order), sort));
         PageRequest pageRequest = PageRequest.of(pageIndex, pageSize, sortExp);
+
         Page<AlertDefine> alertDefinePage = alertDefineService.getAlertDefines(specification, pageRequest);
-        return ResponseEntity.ok(Message.success(alertDefinePage));
+        Page<AlertDefine> alertDefinePages = new PageImpl<>(alertDefinePage.getContent(), pageRequest, alertDefinePage.getTotalElements());
+
+        return ResponseEntity.ok(Message.success(alertDefinePages));
```
tomsun28, line 119: hi why new a pageImpl to replace pre one here

yuluo-yx: The problem is due to the fact that it causes errors during testing, and I also found the same problem in another test class, so I did it this way

but I didn't find out what the exact cause of the problem was, but this does fix the test questions. 😢

yuluo-yx:
This pr relationship is confusing and I will reopen a pr. pls see #2454

PR #2454, opened 2024-08-04
[Improve] add unit test

yuluo-yx:


alerter/src/main/java/org/apache/hertzbeat/alert/controller/AlertDefinesController.java
```diff
@@ -61,7 +64,13 @@ public ResponseEntity<Message<Page<AlertDefine>>> getAlertDefines(
...
             @Parameter(description = "Number of list pages", example = "8") @RequestParam(defaultValue = "8") int pageSize) {
         Page<AlertDefine> alertDefinePage = alertDefineService.getAlertDefines(ids, search, priority, sort, order, pageIndex, pageSize);
-        return ResponseEntity.ok(Message.success(alertDefinePage));
+
+        return  ResponseEntity.ok(Message.success(
+                new PageImpl<>(
+                        alertDefinePage.getContent(),
+                        PageRequest.of(pageIndex, pageSize, Sort.by(new Sort.Order(Sort.Direction.fromString(order), sort))),
+                        alertDefinePage.getTotalElements()))
+        );
```
tomsun28, line 73: Hi, it maybe caused by test cases or other reasons. I feel that we cannot repeat the construction like this.

yuluo-yx:
yep. I think so. I am debug in my local. If has new discover, i will push.
I will mark its draft later.

tomsun28 ***@***.***>于2024年8月5日 周一12:37写道：

yuluo-yx:
@tomsun28 I find question: pls see: https://github.com/spring-projects/spring-data-commons/issues/2987

alerter/src/main/java/org/apache/hertzbeat/alert/controller/AlertDefinesController.java
yuluo-yx: From the solution in the issue, this is one of the ways to solve the problem 🥲

tomsun28:
> @tomsun28 I find question: pls see: [spring-projects/spring-data-commons#2987](https://github.com/spring-projects/spring-data-commons/issues/2987)

👍  it seem this can fix it. https://github.com/spring-projects/spring-data-commons/issues/2987#issuecomment-2072480592 add pageModule in jackson

yuluo-yx:
> > @tomsun28 I find question: pls see: [spring-projects/spring-data-commons#2987](https://github.com/spring-projects/spring-data-commons/issues/2987)
> 
> 👍 it seem this can fix it. [spring-projects/spring-data-commons#2987 (comment)](https://github.com/spring-projects/spring-data-commons/issues/2987#issuecomment-2072480592) add pageModule in jackson

This would be due to additional class configuration, not sure if there are other side effects, I think it's better to use `new PageImpl` to be safe. 🤔

tomsun28:
> This would be due to additional class configuration, not sure if there are other side effects, I think it's better to use `new PageImpl` to be safe. 🤔

hi, the root cause is the serialization problem of the `page`. It is recommended that we solve the problem instead of circumventing it. we has the custom jackson config here like `timemodule`  https://github.com/apache/hertzbeat/blob/master/manager/src/main/java/org/apache/hertzbeat/manager/config/JacksonConfig.java 

```
builder.modules(new SpringDataJacksonConfiguration.PageModule());
```

yuluo-yx:
> > This would be due to additional class configuration, not sure if there are other side effects, I think it's better to use `new PageImpl` to be safe. 🤔
> 
> hi, the root cause is the serialization problem of the `page`. It is recommended that we solve the problem instead of circumventing it. we has the custom jackson config here like `timemodule` https://github.com/apache/hertzbeat/blob/master/manager/src/main/java/org/apache/hertzbeat/manager/config/JacksonConfig.java
> 
> ```
> builder.modules(new SpringDataJacksonConfiguration.PageModule());
> ```

got it, let me try

yuluo-yx:
@tomsun28 hi, This e2e error is probably caused by #2486, I think we can run the ci test again after the #2486 merge.

cc @crossoverJie

alerter/src/main/java/org/apache/hertzbeat/alert/config/JacksonConfiguration.java
```diff
@@ -0,0 +1,41 @@
...
+ * From: <a href="https://github.com/spring-projects/spring-data-commons/issues/2987#issuecomment-2072480592">...</a>
+ */
+
+@Configuration
+public class JacksonConfiguration {
+
+    @Bean
+    public Jackson2ObjectMapperBuilder jackson2ObjectMapperBuilder(SpringDataJacksonConfiguration.PageModule pageModule) {
+
+        return new Jackson2ObjectMapperBuilder().modules(pageModule);
```
tomsun28, line 38: hi，this will conflict with current existed `JacksonConfig` class

tomsun28: how about add `builder.modules(new SpringDataJacksonConfiguration.PageModule());` in `JacksonConfig `  like this. 

image

tomsun28: 🤔 🤔 🤔 🤔 🤔

common/src/main/java/org/apache/hertzbeat/common/config/MVCConfig.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * @author yuluo
+ * @author <a href="mailto:yuluo08290126@gmail.com">yuluo</a>
+ */
+
+@Configuration
+public class MVCConfig implements WebMvcConfigurer {
+
+	/** 默认日期时间格式 */
```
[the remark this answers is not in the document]
yuluo-yx: got it. I am working on this pr.

manager/src/main/resources/application.yml
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 server:
-  port: 1157
+  port: 11570
```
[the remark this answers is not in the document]
yuluo-yx: got

yuluo-yx:
For this pr, the current state is that it may affect all controllers that use Page<T> as a return value, but it does not affect front-end api calls. It only appears in e2e tests and unit tests, so this pr is in pending status, and will be updated again when there is a better solution for serializing Page objects in spring data jpa.

PR #2466, opened 2024-08-04
[improve] : Optimize queues to obtain data and prevent cpu idling

Yanshuming1:
Optimize queues to obtain data and prevent cpu idling


manager/src/main/java/org/apache/hertzbeat/manager/component/alerter/DispatcherAlarm.java
```diff
@@ -135,6 +135,7 @@ public void run() {
                     }
                 } catch (IgnoreException ignored) {
                 } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
```
yuluo-yx, line 138: this code should be in the previous catch block. right?

Yanshuming1: Code block in InterruptedException

yuluo-yx: got it.

tomsun28 (approved):
lgtm!👍

PR #2544, opened 2024-08-17
infra: update actions/checkout version to 4.1.7 

yuluo-yx:
Update the actions/checkout version and use the commit number to control the version.


.github/workflows/backend-build-test.yml
```diff
@@ -42,7 +42,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # 4.1.7
```
tomsun28, line 45: hi recommends using the major version eg `@v2 @v3`, not the commit hash version

yuluo-yx: got it

tomsun28 (approved):

Issue #2556, opened 2024-08-19
[BUG] <local maven test spotless-maven-plugin check error>


tomsun28:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

spotless-maven-plugin check error when run local maven build test, due we has run npm install package in `home`, but the installed package also has the md file. 

```
                    <markdown>
                        <includes>
                            <include>home/**/*.md</include>
                        </includes>
                        <flexmark />
                        <replaceRegex>
                            <name>Markdown Formatter</name>
                            <searchRegex>(^-*\n$)([\s\S]*?)(-+$)</searchRegex>
                            <replacement>---$2---</replacement>
                        </replaceRegex>
                    </markdown>
```

```
[ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.40.0:check (spotless-check) on project hertzbeat: The following files had format violations:
[ERROR]     home/node_modules/is-plain-obj/readme.md
[ERROR]         @@ -4,14 +4,12 @@
[ERROR]          
[ERROR]          An·object·is·plain·if·it's·created·by·either·`{}`,·`new·Object()`,·or·`Object.create(null)`.
[ERROR]          
[ERROR]         -
[ERROR]          ##·Install
[ERROR]          
[ERROR]          ```
[ERROR]          $·npm·install·is-plain-obj
```
how about we use the [markdownlint](https://www.npmjs.com/package/markdownlint) in home to instead of  `spotless` ? , or the github action markdown lint check? https://github.com/DavidAnson/markdownlint @zhangshenghang 

### Environment

```markdown
HertzBeat version(s): master
```

zhangshenghang:
@tomsun28  Can we skip checking the `home/node_modules/ `directory and only check `home/docs` and `home/blog` ，How about this method

tomsun28:
> @tomsun28 Can we skip checking the `home/node_modules/ `directory and only check `home/docs` and `home/blog` ，How about this method

hi this is also OK.  Another benefit of markdownlint is that it supports more markdown standard checks and use [markdownlint-cli2](https://github.com/DavidAnson/markdownlint-cli2) to fixes.

zhangshenghang:
> > @tomsun28 Can we skip checking the `home/node_modules/ `directory and only check `home/docs` and `home/blog` ，How about this method
> 
[...]

PR #2557, opened 2024-08-19
[fixbug] Fixbug doc format temporary process

zhangshenghang:
## What's changed?

#2556 

For the convenience of use, Temporary process


tomsun28 (approved):
👍

PR #2606, opened 2024-08-25
[improve] update thread pool nums policy

tomsun28:
## Checklist

## Add or update API


Calvin979 (approved):

alerter/src/main/java/org/apache/hertzbeat/alert/AlerterWorkerPool.java
```diff
@@ -50,11 +50,11 @@ private void initWorkExecutor() {
                 .setDaemon(true)
                 .setNameFormat("alerter-worker-%d")
                 .build();
-        workerExecutor = new ThreadPoolExecutor(6,
+        workerExecutor = new ThreadPoolExecutor(10,
                 10,
                 10,
                 TimeUnit.SECONDS,
-                new SynchronousQueue<>(),
+                new LinkedBlockingQueue<>(),
```
Calvin979, line 57: It is dangerous here, because the default size of `LinkedBlockingQueue` is `Integer.MAX_VALUE`.
It may cause OOM

Calvin979 (changes requested):

alerter/src/main/java/org/apache/hertzbeat/alert/AlerterWorkerPool.java
tomsun28: Hi, here the threads can be more than 10 at the same time, but the number is unknown(not too much), so linkedqueue is used.

Calvin979 (approved):

PR #2635, opened 2024-08-29
[imporve] Adjustment of parameter verification

MRgenial:
## What's changed?

- Modify the vaild parameter
- Modify relevant unit tests
- Simplify the controller

## Checklist

## Add or update API

### DIFF
--- a/common/src/main/java/org/apache/hertzbeat/common/entity/manager/Tag.java
+++ b/common/src/main/java/org/apache/hertzbeat/common/entity/manager/Tag.java
@@ -29,7 +29,7 @@ import jakarta.persistence.Id;
 import jakarta.persistence.Table;
 import jakarta.validation.constraints.Max;
 import jakarta.validation.constraints.Min;
-import jakarta.validation.constraints.NotNull;
+import jakarta.validation.constraints.NotBlank;
 import java.time.LocalDateTime;
 import java.util.Objects;
 import lombok.AllArgsConstructor;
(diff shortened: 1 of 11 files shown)


Calvin979 (approved):

common/src/main/java/org/apache/hertzbeat/common/entity/manager/Tag.java
```diff
@@ -78,7 +78,7 @@ public class Tag {
             accessMode = READ_WRITE)
     @Min(0)
     @Max(3)
-    private byte type;
+    private Byte type;
```
tomsun28, line 81: Will this change cause null data conversion exceptions?  when type is null in db, but use the `Byte`

Calvin979: > Will this change cause null data conversion exceptions? when type is null in db, but use the `Byte`

I think it is ok. `Byte` is compatible with null but `byte` cannot convert null as a primitive data type

tomsun28: ok 👍

MRgenial: > btw, we need make sure `Byte type` can not null, due here will throws a NullPointerException since null cannot be converted to a primitive value.
> 
> ```
> List<Tag> tags = monitor.getTags().stream().filter(tag -> tag.getType() == (byte) 0).collect(Collectors.toList());
> ```

Okay, i will check it.

PR #2684, opened 2024-09-07
fix: fix e2e unstable

LiuTianyou:
## What's changed?

1. fix e2e unstable

## Checklist

## Add or update API

### DIFF
--- a/collector/src/main/resources/logback-spring.xml
+++ b/collector/src/main/resources/logback-spring.xml
@@ -94,6 +94,7 @@
     <!-- Production environment configuration -->
     <springProfile name="prod">
         <root level="INFO">
+            <appender-ref ref="ConsoleAppender"/>
             <appender-ref ref="SystemOutFileAppender"/>
             <appender-ref ref="ErrOutFileAppender"/>
         </root>
--- a/collector/src/main/resources/logback-spring.xml
+++ b/collector/src/main/resources/logback-spring.xml
@@ -94,7 +94,6 @@
     <!-- Production environment configuration -->
     <springProfile name="prod">
         <root level="INFO">
-            <appender-ref ref="ConsoleAppender"/>
             <appender-ref ref="SystemOutFileAppender"/>
             <appender-ref ref="ErrOutFileAppender"/>
         </root>
--- a/manager/src/main/resources/logback-spring.xml
+++ b/manager/src/main/resources/logback-spring.xml
@@ -94,6 +94,7 @@
     <!-- Production environment configuration -->
     <springProfile name="prod">
         <root level="INFO">
+            <appender-ref ref="ConsoleAppender"/>
             <appender-ref ref="SystemOutFileAppender"/>
             <appender-ref ref="ErrOutFileAppender"/>
         </root>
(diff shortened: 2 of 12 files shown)


collector/src/main/resources/logback-spring.xml
```diff
@@ -94,6 +94,7 @@
     <!-- Production environment configuration -->
     <springProfile name="prod">
         <root level="INFO">
+            <appender-ref ref="ConsoleAppender"/>
```
Aias00, line 97: remove this pls

manager/src/main/resources/logback-spring.xml
```diff
@@ -94,6 +94,7 @@
     <!-- Production environment configuration -->
     <springProfile name="prod">
         <root level="INFO">
+            <appender-ref ref="ConsoleAppender"/>
```
Aias00, line 97: remove this pls

Aias00 (changes requested):

collector/src/main/resources/logback-spring.xml
LiuTianyou: hi，I will fix this later, I am trying to reproduce and fix the e2e instability issue

LinuxSuRen (commented):
[api-testing](https://github.com/LinuxSuRen/api-testing/blob/b6188a0e6831b1af4df52ea63d509d298848ac14/e2e/test-suite-common.yaml#L18) has the similar way to check if the server is ready. But it cannot send a payload to the endpoint.

I'm wondering if there is an API for the health check. Return status code `200` when the server is ready to accept requests.

LiuTianyou:
> [api-testing](https://github.com/LinuxSuRen/api-testing/blob/b6188a0e6831b1af4df52ea63d509d298848ac14/e2e/test-suite-common.yaml#L18) has the similar way to check if the server is ready. But it cannot send a payload to the endpoint.
> 
> I'm wondering if there is an API for the health check. Return status code when the server is ready to accept requests.`200`

@LinuxSuRen Hello, thanks for your reply. I used the login API of hertzbeat to check whether the service is ready, but it seems that the e2e instability is not caused by inaccurate health checks.We have encountered a situation where the API that requires authentication occasionally fails in the test case. Is there a debug mode in api-testing that can output more log information, such as the URL, header, payload of the sent http request, etc. to help us troubleshoot this problem?

LinuxSuRen:
Sure. You can add the log output level flag, such as: `atest run --level debug`

LiuTianyou:
> Sure. You can add the log output level flag, such as: `atest run --level debug`

@LinuxSuRen Thank you. got it

yuluo-yx:
> > Sure. You can add the log output level flag, such as: `atest run --level debug`
> 
> @LinuxSuRen Thank you. got it

Current log message warn level

yuluo-yx (approved):
LGTM

yuluo-yx:
@Aias00 if no other problem, maybe we can merge it, Let's see if the e2e test after that works.

Aias00:
> @Aias00 if no other problem, maybe we can merge it, Let's see if the e2e test after that works.

sure

Aias00 (approved):

Issue #2690, opened 2024-09-08
[Question] should we allow define multi tags with same name?


Aias00:
### Question

should we allow define multi tags with same name?

in my test, i can create multi tags with same name. 

<img width="373" alt="image" src="https://github.com/user-attachments/assets/261c955c-5277-4edc-b31c-09ec73f58bb2">

but after i binded the tags to a monitor

<img width="586" alt="image" src="https://github.com/user-attachments/assets/188187a9-7a10-4ae9-8126-e0024d983a36">

the monitor list just shows one tag

<img width="1274" alt="image" src="https://github.com/user-attachments/assets/79628539-ebc8-40d4-97b7-3444d27399e5">

Aias00:
also, when selecting tags at creating status componment 
<img width="690" alt="image" src="https://github.com/user-attachments/assets/b7104fed-b529-42fb-84f0-de1efeff48e0">

tomsun28:
hi we allow the same tag name with different value, like `logic:xxx, logic:bbb`. but not allow the same name with same value, eg: `logic:aa, logic:aa, logic, logic`

<img width="615" alt="image" src="https://github.com/user-attachments/assets/1c761ce1-2e28-4ac2-ab0a-e803156d6c11">

here seem it is a bug, the tag value does not dispaly.

Aias00:
> hi we allow the same tag name with different value, like `logic:xxx, logic:bbb`. but not allow the same name with same value, eg: `logic:aa, logic:aa, logic, logic`
> 
[...]

PR #2700, opened 2024-09-10
[bugfix] disallow adding tag with same key same value

Aias00:
## What's changed?

disallow adding tag with same key same value
link: https://github.com/apache/hertzbeat/issues/2690

## Checklist

## Add or update API


manager/src/main/java/org/apache/hertzbeat/manager/service/impl/TagServiceImpl.java
```diff
@@ -60,6 +60,10 @@ public class TagServiceImpl implements TagService {
     public void addTags(List<Tag> tags) {
```
tomsun28, line 60: 👍 hi, how about the existed same name and value when modifyTag

Aias00: okay, done

tomsun28 (approved):

Issue #2792, opened 2024-10-29
[Task] 对消息通知中的配置信息进行脱敏处理


shenliehuozhi:
### Description

当前消息通知中的配置信息全都是明文展示，建议对此部分敏感信息进行脱敏处理
![image](https://github.com/user-attachments/assets/19c3aeea-353f-4b98-8a55-59cf7726bc53)

Ceilzcx:
i can do it，please assign to me

ayu-v0:
please assign to me,I want to have a try.I will desensitize sensitive information such as email and mobile phone. @tomsun28

PR #2800, opened 2024-11-02
Desensitize sensitive information

Ceilzcx:
## What's changed?

#2792 

## Checklist

- [ ]  I have read the [Contributing Guide](https://hertzbeat.apache.org/docs/community/code_style_and_quality_guide)
- [ ]  I have written the necessary doc or comment.
- [ ]  I have added the necessary unit tests and all cases have passed.

## Add or update API

- [ ] I have added the necessary [e2e tests](https://github.com/apache/hertzbeat/tree/master/e2e) and all cases have passed.

PR #2848, opened 2024-11-30
[feature] Desensitize sensitive information.

ayu-v0:
## What's changed?
https://github.com/apache/hertzbeat/issues/2792

This PR desensitizes the phone and email, while ensuring that the desensitized data cannot be submitted to the database.
The desensitization rule for a email is to replace it with a "." to make the email illegal and ensure that it cannot be submitted directly to the database.The same goes for cell phone numbers.

Here is a picture of the modification operation：
p1
p2

The following is the information of the database：
p3
p4
## Checklist

## Add or update API


tomsun28:
relate with https://github.com/apache/hertzbeat/pull/2800

ayu-v0:
The merge error is caused by @date and @Description, I've fixed it @yuluo-yx

tomsun28:
hi @ayu-v0 thanks for this pr. 
I reviewed and found that this implementation is not very reasonable. In case user modify the old data, eg: receiver name, phone. Its impact will make people feel confused and think it is a bug.
and another case like send test alram, it will use the serialize data de...@qq.com to send email and failed.
Suggest suspend this issue, investigate the desensitization methods of other platforms later.

tomsun28:
Sorry for rejecting this PR you worked so hard on.

ayu-v0:
> Sorry for rejecting this PR you worked so hard on.

Thank you very much for your review. This PR may be because I considered that the desensitized data is the correct information that the user has verified, so I did not consider that the user might send the email with the desensitized information. Based on this scenario, I wonder if there are any solutions (if I can think of them, haha).

Issue #2847, opened 2024-11-30
[BUG] NPE bug at startup


zqr10159:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

![73a1118c8249596b63afa17091c893d](https://github.com/user-attachments/assets/3abf2c33-7094-442f-b254-ed80ab78b977)

### Environment

```markdown
HertzBeat version(s):1.6.1
```

ayu-v0:
please assign me

ZY945:
I found that getNextCollectMetrics is called in two places, which involves obtaining priorMetrics. Only dispatchMetricsTask calls job. reconstructPriorMetrics();
<img width="1728" alt="image" src="https://github.com/user-attachments/assets/c41e2420-901e-4545-a07a-bef053e99c13">
<img width="1728" alt="image" src="https://github.com/user-attachments/assets/a8e82874-bc45-4580-a425-b9e0e7894c68">
<img width="1728" alt="image" src="https://github.com/user-attachments/assets/b4020b63-8beb-4c96-8484-64f5f0657cca">

ayu-v0:
Perhaps the collection task is initialized the first time and does not need to be initialized later. @ZY945 i think

ayu-v0:
> > Perhaps the collection task is initialized the first time and does not need to be initialized later. @ZY945 i think
> 
[...]

PR #2849, opened 2024-12-01
[bugfix] Fix NullPointerException

ayu-v0:
## What's changed?
https://github.com/apache/hertzbeat/issues/2847

This PR fixes the NullPointerException that occurs when foreach is called when Set is null.

## Checklist

## Add or update API


ayu-v0:
Oh,The merge error is caused by the absence of space after if. I will correct it @yuluo-yx

tomsun28:
hi @ayu-v0 thanks for pr. Due the discuss https://github.com/apache/hertzbeat/issues/2847 , suggest we solved this by find why the getNextCollectMetrics first time return null instead of ignore the NPE.

ayu-v0:
> hi @ayu-v0 thanks for pr. Due the discuss #2847 , suggest we solved this by find why the getNextCollectMetrics first time return null instead of ignore the NPE.

Thank you very much for your review. I shall try to ascertain further the cause of the problem.

ayu-v0:
I found that when setting the monitoring time to 10 seconds, this situation will occur during debugging. This situation is caused by overlapping of the occurrence time of two monitors, which is rare. So it feels like you just need to avoid being empty when the foreach executes🤔(The current thinking is this) @tomsun28 
bf1451198103d2ea5a9245282971841
073ba2ecc6a4b0df9df96e6f3927962
839b93394ca570da2deb0bb56e35318
36089539e42e80ec3facb8269d7f52a
99ee3527713df282731fb5f8ea0e450

tomsun28:
@ayu-v0 Sorry for missing this. 



> I found that when setting the monitoring time to 10 seconds, this situation will occur during debugging. This situation is caused by overlapping of the occurrence time of two monitors, which is rare. So it feels like you just need to avoid being empty when the foreach executes🤔(The current thinking is this) 

+1 @zuobiao-zhou how about this fix way.

zuobiao-zhou:
> @ayu-v0 Sorry for missing this.
> 
> > I found that when setting the monitoring time to 10 seconds, this situation will occur during debugging. This situation is caused by overlapping of the occurrence time of two monitors, which is rare. So it feels like you just need to avoid being empty when the foreach executes🤔(The current thinking is this)
> 
> +1 @zuobiao-zhou how about this fix way.



> @ayu-v0 Sorry for missing this.
> 
> > I found that when setting the monitoring time to 10 seconds, this situation will occur during debugging. This situation is caused by overlapping of the occurrence time of two monitors, which is rare. So it feels like you just need to avoid being empty when the foreach executes🤔(The current thinking is this)
> 
> +1 @zuobiao-zhou how about this fix way.

+1, I only reproduced this issue during debugging, it hasn't occurred during actual use. I believe the problem arises due to prolonged blocking during debugging, which causes the metric collection time to be updated incorrectly, leading to when `job.constructPriorMetrics()` is called next, the condition `.filter(metrics -> (System.currentTimeMillis() >= metrics.getCollectTime() + metrics.getInterval() * 1000))` fails to filter any metrics, leading to an empty priorMetrics. Consequently, `getNextCollectMetrics(null, true)` returns null, resulting in a NullPointerException.

zuobiao-zhou (approved):

Issue #2872, opened 2024-12-12
[Question] <title>连接ClichHouse集群Collect Timeout No Response


Figure43:
### Question

hertzbeat 版本：v1.6.1
clickhouse集群版本：22.12.3.5

使用Kubernetes 部署hertzbeat和使用docker部署hertzbeat都无法连接到Clickhouse集群，但是使用1.6.0版本的hertzbeat可以连接
![image](https://github.com/user-attachments/assets/c1df8b85-0408-438d-a5be-ea29e09e9519)
![image](https://github.com/user-attachments/assets/f10aa9a1-0602-45fa-9ac2-64265c6688b1)

LiuTianyou:
@Figure43  hi, other changes have affected this. You can change line 107 of the clickhouse monitoring template from
 `- name: ping_available` to 
`- name: summary` and it will work properly.

PR #2874, opened 2024-12-12
[Fix] fix clickhouse monitor

LiuTianyou:
## What's changed?

1. Fix the problem that clickhouse monitoring cannot be used normally #2872 

## Checklist

## Add or update API


Calvin979 (approved):

tomsun28:
A bit confused🐶🐶, why change the ping_available -> summary then it works

Calvin979:
> A bit confused🐶🐶, why change the ping_available -> summary then it works

Looks like it is caused by #2266

Calvin979 (changes requested):
I think it is better to change code in `HttpCollectImpl`

tomsun28:
> I think it is better to change code in `HttpCollectImpl`

Agree, metrics are not recommended to be hardcoded in code

yuluo-yx (dismissed):

LiuTianyou:
> I think it is better to change code in `HttpCollectImpl`

hi, #2266 In order for http protocol collection to support obtaining data from response headers, should we add a parsing type as `response_header` and obtain the header data corresponding to the metrics name?

Calvin979:
> > I think it is better to change code in `HttpCollectImpl`
> 
> hi, #2266 In order for http protocol collection to support obtaining data from response headers, should we add a parsing type as `response_header` and obtain the header data corresponding to the metrics name?

Yes, it is better than hard code

Calvin979 (approved):
LGTM

Issue #2885, opened 2024-12-19
[Question] <title>Some optimization suggestions for Kafka client monitoring metrics.


doveLin0818:
### Question

Currently, Hertzbeat offers two methods for monitoring Kafka: one is based on the JMX protocol, and the other uses Kafka clients. For Kafka clients, I have two optimization suggestions:

1. In the "Topic Offset" module, the second column is intended to represent the partition number corresponding to the topic, not the number of partitions, as shown in the figure below. This can cause confusion for Chinese users. I suggest changing the title to "分区号" in the Chinese display. This would require modifying the monitoring template.
<img width="735" alt="image" src="https://github.com/user-attachments/assets/be0e1b39-30a8-4140-bdee-4e3370c64e97" />

2. Currently, Kafka client metrics are limited. I recommend adding a new module called "Consumer Group Status," which would include metrics such as consumer group ID, subscribed topic names, partition offsets, total consumption lag, and other related fields.

If necessary, I am willing to assist in implementing these changes.

tomsun28:
+1👍 welcome to optimize this.

PR #2887, opened 2024-12-19
[collector]feature:Add monitoring metrics for consumer groups in Kafka client

doveLin0818:
Issue：https://github.com/apache/hertzbeat/issues/2885

## What's changed?
1.The Kafka client monitoring metrics have been extended with a consumer information module, including consumer group ID, number of consumer group members, topics subscribed by the consumer group, partition offset, and lag offset metrics and each metric has been tested.

2.The ambiguity in the Chinese description of the partition count field in the topic_offset module has been fixed.

3.The hertzbeat community documentation has also been updated accordingly.

The example diagram is shown below:
image

## Checklist

- [✅]  I have read the [Contributing Guide](https://hertzbeat.apache.org/docs/community/code_style_and_quality_guide)
- [✅]  I have written the necessary doc or comment.
- []  I have added the necessary unit tests and all cases have passed.

## Add or update API


zhangshenghang (changes requested):
image

thanks @doveLin0818 , It is recommended to change Offset of Each Partition to [0,10,2] for easier understanding

doveLin0818:
> 图像
> 谢谢@doveLin0818，建议将每个Partition的Offset改为[0,10,2]，这样比较容易理解

#2887 @zhangshenghang "Hello, yeah，This way is easier to understand. I have now modified the format as shown in the image below."
image

hertzbeat-manager/src/main/resources/define/app-kafka_client.yml
Rancho-7, line 1: Thanks for improvement!

In the `consumer_detail` section, would it be necessary to associate the specific partition number, offset, and lag offset for each partition? 

Because, in the case where multiple partitions have lag, how could we identify which lag offset corresponds to which partition?

doveLin0818: For the offsets of each partition, they are displayed sequentially here. You can view the offsets of each partition in the 'Topic Offsets' module. The lag shown here refers to the total offset lag at the consumer group level, not at the partition granularity. If necessary, the 'Offsets per Partition' can be displayed in a format like [0:69, 1:280]

Rancho-7: > the 'Offsets per Partition' can be displayed in a format like [0:69, 1:280]

That makes sense.Thanks.

zhangshenghang (approved):
LGTM!

PR #2901, opened 2024-12-25
[Imporve] Support Kafka internal topic configuration

zhangshenghang:
## What's changed? 

Kafka internal topics do not need to be displayed in most cases, as they take up too much space. Because he has 50 partitions.
We use configuration items to set whether to monitor internal topics.

image

image


tomsun28 (approved):
lgtm.

PR #3011, opened 2025-01-22
[MINOR UPDATE] always specify the char encoding in getBytes

pjfanning:
* if you don't specify the char encoding, then it is up to the JVM config on the machine
* this means different users can see different results


tomsun28 (approved):
LGTM 👍

PR #3015, opened 2025-01-23
[issue-2998] remove invalid check in isValidLabelValue

### LINKED ISSUE #2998: [Question] isValidLabelValue doesn't look right (to me)
### Question

https://github.com/apache/hertzbeat/blob/fd7e1b0e056e949f9fb5d7a7e2e4e487f7e95b61/hertzbeat-collector/hertzbeat-collector-basic/src/main/java/org/apache/hertzbeat/collector/collect/http/promethus/exporter/ExporterParser.java#L421

```
    /**
     * Checks if a string is valid UTF-8 encoded
     *
     * @param s label value
     * @return true/false
     */
    private static boolean isValidLabelValue(String s) {
        return s != null && s.equals(new String(s.getBytes(StandardCharsets.UTF_8)));
    }
```

The String `s` is already converted to chars - so it is too late to work out if it originates from UTF-8 encoded bytes.

A proper check would take the binary data and check it, not convert it to a String first which presumably doesn't even check what the encoding is.

One approach might be:
https://stackoverflow.com/questions/46856381/how-to-check-if-byte-array-is-valid-utf-8-string

A similar issue seems to be in https://github.com/apache/hertzbeat/blob/fd7e1b0e056e949f9fb5d7a7e2e4e487f7e95b61/hertzbeat-collector/hertzbeat-collector-basic/src/main/java/org/apache/hertzbeat/collector/collect/prometheus/parser/OnlineParser.java#L243

and
https://github.com/apache/hertzbeat/blob/fd7e1b0e056e949f9fb5d7a7e2e4e487f7e95b61/hertzbeat-collector/hertzbeat-collector-basic/src/main/java/org/apache/hertzbeat/collector/collect/prometheus/parser/TextParser.java#L302

pjfanning:
## What's changed?

fixes #2998 

## Checklist

## Add or update API


tomsun28 (approved):
lgtm

PR #3072, opened 2025-02-12
[bugfix](web-app): update app menu visibility on monitor creation

zqr10159:
- Add updateLocalAppState method to update app menu visibility
- Call updateLocalAppState when creating a new monitor

## Checklist

## Add or update API

### DIFF
--- a/web-app/src/app/routes/monitor/monitor-new/monitor-new.component.ts
+++ b/web-app/src/app/routes/monitor/monitor-new/monitor-new.component.ts
@@ -19,7 +19,7 @@
 
 import { Component, Inject, OnInit } from '@angular/core';
 import { ActivatedRoute, ParamMap, Router } from '@angular/router';
-import { I18NService } from '@core';
+import { I18NService, StartupService } from '@core';
 import { ALAIN_I18N_TOKEN, TitleService } from '@delon/theme';
 import { NzNotificationService } from 'ng-zorro-antd/notification';
 import { switchMap } from 'rxjs/operators';
@@ -32,6 +32,7 @@ import { Param } from '../../../pojo/Param';
 import { ParamDefine } from '../../../pojo/ParamDefine';
 import { AppDefineService } from '../../../service/app-define.service';
 import { CollectorService } from '../../../service/collector.service';
+import { GeneralConfigService } from '../../../service/general-config.service';
 import { MonitorService } from '../../../service/monitor.service';
 import { generateReadableRandomString } from '../../../shared/utils/common-util';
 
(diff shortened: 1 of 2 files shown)


web-app/src/app/routes/monitor/monitor-new/monitor-new.component.ts
```diff
@@ -168,13 +171,19 @@ export class MonitorNewComponent implements OnInit {
...
-          this.router.navigateByUrl(`/monitors?app=${info.monitor.app}`);
+          this.configService.updateAppTemplateConfig({ hide: false }, info.monitor.app).subscribe(() => {
+            this.startUpSvc.loadConfigResourceViaHttp().subscribe(() => {
+              this.isSpinning = false;
+              this.notifySvc.success(this.i18nSvc.fanyi('monitor.new.success'), '');
+              this.router.navigateByUrl(`/monitors?app=${info.monitor.app}`);
+            });
+          });
         } else {
+          this.isSpinning = false;
```
tomsun28, line 186: hi, why need updateAppTemplateConfig when add a new monitor, the `this.startUpSvc.loadConfigResourceViaHttp()` is a very time-consuming initialization function operation

zqr10159: When a user adds a new type of monitor, he should want to see it in the left navigation bar right after adding the monitor. So I think a refresh function should be added here. There seems to be a need to add a judgment here: is it the first time a certain type of monitor is added?

tomsun28: I think the display of the main menu and the monitoring types added by the user should not be strictly bound. 
Users can also view the added monitors by clicking on the monitoring center. The monitoring type menu in the main menu is just a shortcut. For example, in this case, it is possible that the user does not want to display the monitoring type in the main menu for simplicity. It is recommended to let the user decide whether to display this monitoring type in the main menu by editing monitoring template here instead of auto.
image

zqr10159: I think the button to show the menu is too hidden and there is no hint of an interface that would confuse the user if used for the first time: why is there no category in the left menu? This is not user friendly for novice users, is there another way to make a good hint? I suggest discussing this on the mailing list.

tomsun28: OK, 👀 we can discuss it on the mailing list or [github discussions](https://github.com/apache/hertzbeat/discussions)

PR #3108, opened 2025-02-25
[bugfix] Fix http header being incorrectly encoded.

yunfan24:
## What's changed?

#3086  
Remove special character replacement for HTTP headers and paths for the following reasons:  

1. For HTTP headers, certain special headers (such as Authorization, Content-Type, etc.) may fail authentication or cause content type recognition errors if encoded. For example, Basic authentication must strictly follow the format: `Basic <Base64-encoded-credentials>`. If `Basic` is encoded as `Basic%20`, the server will not recognize it as a Basic authentication header. Since headers are generally not processed this way, I did not add a switch for encoding headers.  

2. Encoding the path may mistakenly convert meaningful special characters, such as the path separator `/`, which could affect the correctness of the request.  

By preserving the original values and allowing users to decide whether to encode specific parts manually, we provide greater customization and flexibility.

## Checklist

## Add or update API


tomsun28:
hi, we can't simply remove it, because there are other cases where an encoded URI is required.
It is recommended that we provide a switch to select whether to encode, which is enabled by default. Users can configure whether to use encode on the monitor edit page.

yunfan24:
> hi, we can't simply remove it, because there are other cases where an encoded URI is required. It is recommended that we provide a switch to select whether to encode, which is enabled by default. Users can configure whether to use encode on the monitor edit page.

Hello, I tested multiple encoding APIs, and perhaps we can use `org.springframework.web.util.UriUtils.encodePath()` to encode the path. This method encodes characters like Chinese characters or spaces but won't encode meaningful characters such as "/". For example:

```
/api/hello world/你好/_aaabbb/ccc
```

After encoding, it becomes:

```
/api/hello%20world/%E4%BD%A0%E5%A5%BD/_aaabbb/ccc
```

Maybe we can replace the usage of `org.apache.hertzbeat.collector.util.CollectUtil#replaceUriSpecialChar` with `UriUtils.encodePath()` in the project.

Regarding enabling encoding by default, I believe it would be more consistent with user habits, as browsers and tools like Postman automatically encode user input. However, we may need to manually edit some monitors when upgrading from the old version, because currently we only encode spaces in the path. Users might have created some HTTP APIs with special characters and manually encoded them. Enabling encoding by default in the new version might cause double encoding. This should be documented, or we need to find other ways to avoid compatibility issues.

Additionally, regarding the switch for URL encoding, I need to confirm one more thing: if the switch is turned off, should both relative paths (e.g., `/aaa/bbb`) and parameters (e.g., `?name=mike&age=12`) stop being encoded? In the current implementation, `requestBuilder.addParameter` automatically encodes parameters by default. If we decide not to encode parameters when the switch is off, this part of the code will also need to be changed.
https://github.com/apache/hertzbeat/blob/36889746100fc208869e2b1c1600d1afcc489091/hertzbeat-collector/hertzbeat-collector-basic/src/main/java/org/apache/hertzbeat/collector/collect/http/HttpCollectImpl.java#L510-L518

tomsun28:
> Additionally, regarding the switch for URL encoding, I need to confirm one more thing: if the switch is turned off, should both relative paths (e.g., `/aaa/bbb`) and parameters (e.g., `?name=mike&age=12`) stop being encoded? In the current implementation, `requestBuilder.addParameter` automatically encodes parameters by default. If we decide not to encode parameters when the switch is off, this part of the code will also need to be changed.

hi, agree, 👍 i think the encoding switch button should control the url, relative path and parameters encoding.

yunfan24:
Modified completed

yunfan24:
Hi, should we allow directly concatenating parameters in the path? It seems that some places in the project use this method, directly concatenating parameters in the path for passing. This has caused unit test failures. @tomsun28 
https://github.com/apache/hertzbeat/blob/268e11102418c014af70cc55abc8ce06533295a7/hertzbeat-manager/src/main/resources/define/app-docker.yml#L235-L248

tomsun28:
> Hi, should we allow directly concatenating parameters in the path?

hi yes, we allow this case.

yunfan24:
> > Hi, should we allow directly concatenating parameters in the path?
> 
> hi yes, we allow this case.

Okay, I will make the modifications.

tomsun28 (approved):
LGTM!

Issue #3184, opened 2025-03-28
[BUG] <title>1.7.0版本阈值规则 关联设备超过46条后无法保存


gaows20:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

1.7.0版本新建或修改阈值规则，单条规则关联设备超过46条后无法保存
错误提示：
修改失败!
Size.alertDefine.expr:个数必须在0和2048之间

### Environment

```markdown
HertzBeat version(s):
```

bigcyy:
I have reproduced this bug. The reason is that the backend receives `expr` as a string and validates it by checking that its length does not exceed 2048 characters. Perhaps we need to find a better way to handle this.

bigcyy:
note：This modification is expected to support an alert being associated with approximately 100 monitors.

Does this meet your requirements?

gaows20:
我设备数量较多，对我而言不够，大约需要600个，如果只能是6144的话我多添加几条规则就是了，但是如果关联设备能支持按照label选择的话就更好了。

Issue #3185, opened 2025-03-28
[Task] recommended to support label selection when monitoring threshold rules


gaows20:
### Description

1.7.0版本 阈值规则关联监控时建议支持按label选择，设备太多时挨个选择太繁琐。

### Task List

1： 1.7.0版本阈值规则关联监控时建议支持按label选择

LL-LIN:
You can refer to:
https://github.com/apache/hertzbeat/pull/3161

tomsun28:
> Please assign it to me, I will complete it.

welcome.

tomsun28:
> Please assign it to me, I will complete it.

hi, Is the plan to replace the previous one with labels, or will both methods exist? 

Suggest also put the label association in the `Alert Threshold Association` modal box. 
<img width="1272" alt="Image" src="https://github.com/user-attachments/assets/e0c05d2a-f7a4-45fb-bb54-3886362c4835" />

bigcyy:
when selecting monitoring, you can filter directly through tags.

https://github.com/user-attachments/assets/1e59212e-2ecf-46cb-b3c3-a7e5fd1323a6

tomsun28:
@gaows20 Hi, is this what you mean? I thought it was label matching, not label filtering. The label matching is to directly associate the threshold rule with the corresponding monitored label by configuring the label, and finally calculate the threshold. Similar to the expression generated at the end, it contains label match expr like `contains(__labels__, 'banana')`, eg: `equals(__app__,"website") && equals(__metrics__,"summary") && contains(__labels__, 'banana')`. This situation can also simplify the problem of having thousands of monitoring points to be selected.

gaows20:
我目前需要的是:标签匹配。某个标签下有600个设备，把这个标签下所有设备都放在一个阈值规则。这和我提的那个超过46条添加失败差不多属于一个问题。当然我感觉这个标签过滤也是一个非常好的功能。以后可能会用到。

bigcyy:
Thank you, Tom and gaows20, for your clarification. I will add label filtering and label matching.

tomsun28:
👍 so according to the conversation, seems this is no need to update expr length, or we can update `@Size(max = 2048)
    @Column(length = 2048)` to `4096` and remove the update sql(Flyway sql may sometimes fail to start in different environments. Try not to use it if it is not necessary.) https://github.com/apache/hertzbeat/pull/3198

bigcyy:
[...]

PR #3198, opened 2025-03-30
[bugfix] increase length limit for alert rule expressions to support large device lists

bigcyy:
## What's changed?

#3184 

This PR modifies the database schema to use larger text field types appropriate for each supported database:

- MySQL: Changed to LONGTEXT type
- PostgreSQL: Using TEXT type
- H2: Using CLOB type

The implementation adds the JPA `@Lob` annotation to the expr field 

Testing

- Verified that rules with 1000 devices can be successfully created and saved
- Tested across all supported database types

## Checklist

## Add or update API


gaows20:
👍啥时候合并到主线，进nightly哇

tomsun28:
hi, maybe we need to consider whether it is appropriate to use `LONGTEXT`(max 4 GB) to store expr string.
In most cases, expr will be very short. If you encounter a large number of monitors that need to be manually bound, consider using label binding. Manual binding to select thousands of monitors is not very convenient. 
refer https://github.com/apache/hertzbeat/issues/3185

PR #3232, opened 2025-04-10
[doc] code style check

a-little-fool:
## What's changed?

Here, it asks to download markdownlint-cli2, but then asks to run markdownlint, which will result in a "command not found" error. The correct way should be to run markdownlint-cli2.

截屏2025-04-10 23 03 45

截屏2025-04-10 23 03 13


tomsun28 (approved):
👍

Issue #3239, opened 2025-04-13
[BUG] CollectorJobScheduler: Job Cache Entries Not Removed When Canceling Async Jobs


bigcyy:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

In the `CollectorJobScheduler` class, when adding an async collection job via `addAsyncCollectJob()`, the job is stored in the `jobContentCache` map. However, when canceling a job using `cancelAsyncCollectJob()`, the job is removed from collector nodes but not from the `jobContentCache`.

### Expected Behavior

The `cancelAsyncCollectJob()` method should remove the job entry from the `jobContentCache` map, similar to how it's done in the `updateAsyncCollectJob()` methods.

### Steps To Reproduce

1.Call addAsyncCollectJob() to add a new collection job
2.Call cancelAsyncCollectJob() to cancel the job
3.Observe that the entry remains in the jobContentCache map
4.Over time, as many jobs are created and canceled, the jobContentCache will grow unbounded and consume memory

### Environment

```markdown
HertzBeat version(s):
```

PR #3240, opened 2025-04-13
[bugfix] Added jobContentCache.remove(jobId) to cancelAsyncCollectJob

bigcyy:
## What's changed?

#3239 

Added jobContentCache.remove(jobId) to cancelAsyncCollectJob method to prevent memory leak. Jobs were being added to the cache but not removed when canceled, causing the map to grow unbounded over time.

## Checklist

## Add or update API


tomsun28:
👍👍 Will this problem also trigger that when redispatch jobs, the deleted monitoring tasks will be assigned to the new node to continue monitoring before?

bigcyy:
> 👍👍 Will this problem also trigger that when redispatch jobs, the deleted monitoring tasks will be assigned to the new node to continue monitoring before?

It doesn't seem like it will, because all nodes are traversed here and the deletion is executed. However, the memory leak issue still exists — even if the tasks are removed from all nodes, if they are not deleted from jobContentCache, these objects will still occupy memory and cannot be garbage collected. In a long-running system, this can lead to continuous memory growth.。

image

tomsun28 (approved):
👍 LGTM

Issue #3220, opened 2025-04-07
[BUG] There is a bug in RealTimeAlertCalculator during threshold rule filtering


bigcyy:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

There is a bug in `RealTimeAlertCalculator` during threshold rule filtering. Specifically, when an alert expression is associated with only one instance (e.g., `redis1`), unassociated instances (e.g., `redis2`) are **not properly filtered out** and still get passed to the Jexl engine for evaluation.

## Root Cause

The issue is caused by a mismatch between the format of the expression generated by the frontend and the regex used by the backend to detect instance.

**Frontend behavior:**

The frontend generates expressions like:

```text
equals(__instance__, "99992324230")
```

Note the **space between the comma and the opening quote**.

**Backend behavior:**

The backend regex used for matching:

```java
equals\(__instance__,"(\d+)"\)
```

does **not account for optional whitespace**, which causes `.find()` to fail. As a result, the fallback condition is triggered:

```java
return true;
```

This leads to **all instances being accepted**, regardless of whether they were explicitly bound.

![Image](https://github.com/user-attachments/assets/51cc13d0-0278-4e5a-8668-1da814a15dbe)

![Image](https://github.com/user-attachments/assets/cda76e02-e5ed-4e2e-9063-bf1afff43767)

![Image](https://github.com/user-attachments/assets/636e86e8-dd59-4b28-94ae-8dba37b8675c)

### Expected Behavior

Only instances explicitly matched by the expression should be evaluated.  
Unbound instances (i.e., those not matching any condition like equals(__instance__, "982923424230")) should be filtered out and not passed to the Jexl engine, even if the expression contains whitespace after the comma.

### Steps To Reproduce

1. Create an alert rule associated with only one instance (e.g., redis1).
2. Let the frontend generate the expression: equals(__instance__, "982923424230").
3. Observe that unassociated instances (e.g., redis2, redis3) are still processed by the alert calculator.

### Environment

```markdown
HertzBeat version(s):
```

bigcyy:
Although this does not affect the correctness of alert results, it introduces a performance issue. For example, in a scenario with 50 Redis instances where only one is configured for alerting, the other 49 unassociated instances still get evaluated unnecessarily.

I will try to fix it by updating the regex to allow optional whitespace after the comma:  
```
equals\\(__labels__,\\s*\"([^\"]+)\"\\)
```
[...]

PR #3269, opened 2025-04-18
[bugfix] correct instance filtering regex in RealTimeAlertCalculator

bigcyy:
## What's changed?

#3220 

This PR fixes a bug where `RealTimeAlertCalculator` failed to properly filter unbound instances during threshold rule evaluation. The issue was caused by the backend regex not accounting for optional whitespace in alert expressions generated by the frontend.

Log-based verification was used to confirm correctness of the fix:

- Logs before and after filtering were printed.
- Only the bound instance was allowed to pass the filter, while unbound ones were filtered out as expected.

Sample log output:

```txt
filter before expr: equals(__app__,"redis") && equals(__available__,"down") && equals(__instance__, "501045327364864"), instance :501786917278464  
filter before expr: equals(__app__,"redis") && equals(__available__,"down") && equals(__instance__, "501045327364864"), instance :501045327364864  
filter after expr: equals(__app__,"redis") && equals(__available__,"down") && equals(__instance__, "501045327364864"), instance :501045327364864
```
Note:

The RealTimeAlertCalculator class is a passive-invocation component with only a public constructor. It registers tasks and starts threads during construction, while core logic resides in private methods triggered asynchronously. This makes unit testing difficult without reflection or design adjustments. 

## Checklist

## Add or update API


tomsun28:
LGTM! 👍 It would be better if could add some unit test cases.

tomsun28 (approved):

PR #3273, opened 2025-04-19
[improve] Add jdbc common collect e2e code (#3273)

Cyanty:
## What's changed?

Benefit from testcontainers' rich jdbc driver support (SPI implementation), add jdbc common collect e2e code (postgresql/mysql supported for now)

## Checklist

- [x]  I have read the [Contributing Guide](https://hertzbeat.apache.org/docs/community/code_style_and_quality_guide)
- [x]  I have written the necessary doc or comment.
- [ ]  I have added the necessary unit tests and all cases have passed.

## Add or update API

- [ ] I have added the necessary [e2e tests](https://github.com/apache/hertzbeat/tree/master/e2e) and all cases have passed.

Issue #2096, opened 2024-06-19
[BUG] monitoring k8s error


MrFrankBolt:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

按照文档生成了对应的k8s认证token，然后监控集群，发现一直处于宕机状态，观察后端log发现如下：
![e3f7d3b3c217f7810ff18c770f2ec4ac](https://github.com/apache/hertzbeat/assets/120542351/34752a3e-f140-4891-ab59-f99feb14868c)

### Expected Behavior

希望可以正常监控k8s，采集对应的指标

### Steps To Reproduce

1. 使用的是1.6.0镜像版本，之前是在1.4.4版本，按照文档进行了升级；
2. 配置只做了数据源相关的更改，其他都是原始默认，k8s采集yaml也是使用的默认；
3. 使用docker启动；
4. 查看k8s监控结果，出现上述描述异常；

### Environment

```markdown
HertzBeat version(s):v1.6.0
```

### Debug logs

2024-06-19 17:01:09.893 [339555282159616-kubernetes-nodes-9839] ERROR org.apache.hertzbeat.collector.dispatch.WorkerPool Line:48  - Thread Name 339555282159616-kubernetes-nodes-9839 : Character K is neither a decimal digit number, decimal point, nor "e" notation exponential mark.
java.lang.NumberFormatException: Character K is neither a decimal digit number, decimal point, nor "e" notation exponential mark.
        at java.base/java.math.BigDecimal.<init>(BigDecimal.java:582)
        at java.base/java.math.BigDecimal.<init>(BigDecimal.java:467)
        at java.base/java.math.BigDecimal.<init>(BigDecimal.java:896)
        at org.apache.hertzbeat.collector.dispatch.unit.impl.DataSizeConvert.convert(DataSizeConvert.java:46)
        at org.apache.hertzbeat.collector.dispatch.MetricsCollect.calculateFields(MetricsCollect.java:310)
        at org.apache.hertzbeat.collector.dispatch.MetricsCollect.run(MetricsCollect.java:176)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
        at java.base/java.lang.Thread.run(Thread.java:833)

ncuhe:
[...]

PR #3283, opened 2025-04-22
[improve] Add E2E tests for Redis collector

KevinLLF:
## What's changed?

enable RedisCommonCollect E2E tests via Testcontainers

## Checklist

## Add or update API

### DIFF
--- a/hertzbeat-e2e/hertzbeat-collector-basic-e2e/src/test/java/org/apache/hertzbeat/collector/collect/basic/redis/RedisCommonCollectE2eTest.java
+++ b/hertzbeat-e2e/hertzbeat-collector-basic-e2e/src/test/java/org/apache/hertzbeat/collector/collect/basic/redis/RedisCommonCollectE2eTest.java
@@ -1,0 +1,126 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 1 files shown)


tomsun28:
hi It is not recommended to put e2e testing in unit test cases. We have an independent e2e testing environment module. refer: https://github.com/apache/hertzbeat/pull/3280 hi @zhangshenghang how about this?

tomsun28:
https://github.com/apache/hertzbeat/pull/3273 or https://github.com/apache/hertzbeat/pull/3280 may we need to decide which e2e test method to use, or both. @zhangshenghang @LiuTianyou @zqr10159

zhangshenghang:
> #3273 or #3280 may we need to decide which e2e test method to use, or both. @zhangshenghang @LiuTianyou @zqr10159

I think both can coexist.

In #3280, some general interface monitoring can be done from the overall service, which is necessary, but it cannot support too many containers to start and test at the same time, and resources will be insufficient.
In #3273, each component can be monitored, because testcontainers will quickly release resources after execution.

LiuTianyou:
> https://github.com/apache/hertzbeat/pull/3273 or https://github.com/apache/hertzbeat/pull/3280 may we need to decide which e2e test method to use, or both. @zhangshenghang @LiuTianyou @zqr10159 

hi, I also think we can use both methods. for general collection capabilities, we can use #3273 to ensure the correctness of the collection function. for a specific monitoring template, use #3280 to ensure the availability of the newly added template and that new code modifications will not affect the previously released monitoring template. using e2e like #3280 can prevent problems similar to #2872 and #2096 from occurring again.

PR #3365, opened 2025-05-17
[improve] improve url validation for GotifyAlertNotifyHandlerImpl

Aias00:
[improve] improve url validation for GotifyAlertNotifyHandlerImpl


hertzbeat-alerter/src/main/java/org/apache/hertzbeat/alert/notice/impl/GotifyAlertNotifyHandlerImpl.java
```diff
@@ -104,4 +109,19 @@ public static class ClientDisplay {
...
+     * Validate the webhook URL against a whitelist of allowed base URLs.
+     *
+     * @param url the webhook URL to validate
+     * @return true if the URL is valid, false otherwise
+     */
+    private boolean isValidWebhookUrl(String url) {
+        // Define a whitelist of allowed base URLs
+        List<String> allowedBaseUrls = List.of(
+                "https://trusted-gotify-server.com",
+                "https://another-trusted-server.com"
```
tomsun28, line 123: Gotify is self-hosted server, can not valid url.

PR #3366, opened 2025-05-17
[improve] improve url validation for FlyBookAlertNotifyHandlerImpl

Aias00:
[improve] improve url validation for FlyBookAlertNotifyHandlerImpl


hertzbeat-alerter/src/main/java/org/apache/hertzbeat/alert/notice/impl/FlyBookAlertNotifyHandlerImpl.java
```diff
@@ -202,4 +210,14 @@ private String createLarkMessage(String userId, String notificationContent, byte
     public byte type() {
         return 6;
     }
+    
+    private boolean isValidBaseUrl(String baseUrl) {
+        // Ensure the base URL is a trusted, fixed URL
+        return baseUrl != null && baseUrl.startsWith("https://trusted-domain.com/");
+    }
```
tomsun28, line 217: flybook url is not trusted-domain, and else,there will be a deployment of private Feishu, and the URL domain name is not fixed.

PR #3371, opened 2025-05-17
[improve] improve url validation for MetricsDataController

Aias00:
[improve] improve url validation for MetricsDataController


hertzbeat-warehouse/src/main/java/org/apache/hertzbeat/warehouse/store/history/vm/VictoriaMetricsDataStorage.java
```diff
@@ -532,4 +537,9 @@ public static final class VictoriaMetricsContent {
          */
         private Long[] timestamps;
     }
+
+    private boolean isTrustedBaseUrl(String url) {
+        // Define a trusted base URL
+        return url.equals("http://trusted-victoriametrics-server.com");
```
tomsun28, line 543: Can not fix like this, victormetrics is self-hosted url.

Aias00: how about set up a properties for user

PR #3406, opened 2025-05-27
[feat] add __script__ support for SQL and PromQL expressions

bigcyy:
## What's changed?

#3397 
This PR introduces support for parsing and handling __script__("<script>") style expressions within periodic threshold rules. The motivation is to enhance flexibility by supporting SQL-style expressions in addition to existing PromQL expressions. The update includes:

- New lexical rule: __script__(...) syntax is now recognized by the lexer.
- New syntax rule: supports quoted string content for embedded SQL or PromQL.
- Backward compatibility: original PromQL queries without __script__ are still supported and prioritized as default.
- Update unit test.

## Checklist

- [X]  I have added the necessary unit tests and all cases have passed.

## Add or update API


tomsun28:
cc @MasamiYui hi  i found we have refact by this discussion https://github.com/apache/hertzbeat/discussions/3205, how about this current pr?

MasamiYui:
Hi, I think embedded queries are appropriate，but named script() here is maybe not suitable. scripts focus more on action choreography, and the implementation here is more like a query function that requires return values.
To support expressions like `select count(num) from table > 10` maybe import function named like sql_query() or sql()?

bigcyy:
> and if can support [#3205 (comment)](https://github.com/apache/hertzbeat/discussions/3205#discussioncomment-13295280) this, it would be better.

I will keep both, because even though `select` syntax is supported, due to differences in database dialects, we still need syntax like `sql("...")` for compatibility.

PR #3410, opened 2025-05-29
[feat] support sql("...") and promql("...") expressions including SQL condition parsing

bigcyy:
## What's changed?

#3205 
- This PR enhances threshold rule expression handling by introducing support for embedded SQL and PromQL queries via `sql("...")` and `promql("...")` syntax. 
- It also extends SQL parsing capabilities to support conditional logic directly within SQL expressions (e.g., `SELECT count(*) FROM table WHERE value > 10`).
- Added tests for valid and invalid SQL expressions.

## Checklist

- [X]  I have added the necessary unit tests and all cases have passed.

## Add or update API


MasamiYui:
hi,I have two questions:
1.If use the local grammar file parse sql, maybe unable to use certain advanced queries or dialects.
2.I noticed you modified the cache logic,is there any consideration here.

bigcyy:
> hi,I have two questions:
> 1.If use the local grammar file parse sql, maybe unable to use some advanced queries or dialects.
> 2.I noticed you modified the cache logic,is there any consideration here

Hi, here are my responses to your questions:

1. Regarding SQL parsing:
You're absolutely right — using a local grammar file may not support certain advanced queries or database-specific dialects. That’s why I’ve decided to retain both sql("") and promql("") styles, to ensure full compatibility with different SQL engines and PromQL syntax.

2. About the cache logic modification:
Previously, if I needed to access the tokens (to extract the original SQL/PromQL string), I had to retrieve them from the cache again, which was redundant. So I refactored the code: now it first obtains the tokens, then generates the parse tree. This improves clarity and avoids repeated cache lookups.

Let me know if you have any suggestions or concerns.

MasamiYui:
> 2\. About the cache logic modification:
>     Previously, if I needed to access the tokens (to extract the original SQL/PromQL string), I had to retrieve them from the cache again, which was redundant. So I refactored the code: now it first obtains the tokens, then generates the parse tree. This improves clarity and avoids repeated cache lookups.

how about use `ctx.selectSql().getText()` replace  `tokens.getText(ctx.selectSql())`

bigcyy:
> > 2. About the cache logic modification:
> > Previously, if I needed to access the tokens (to extract the original SQL/PromQL string), I had to retrieve them from the cache again, which was redundant. So I refactored the code: now it first obtains the tokens, then generates the parse tree. This improves clarity and avoids repeated cache lookups.
> 
> how about use `ctx.selectSql().getText()` replace `tokens.getText(ctx.selectSql())`

Hi, I tried using ctx.selectSql().getText(), but it doesn't return the full SQL string with spaces — I only get something like `select*fromtable`. However, when I use tokens.getText(ctx.selectSql()), it correctly returns `select * from table`.

tomsun28 (approved):
lgtm

PR #3420, opened 2025-05-31
[bugfix] CommonDispatcher#dispatchCollectData may get `NullPointerException` if not a service discovery job

Rancho-7:
## What's changed?
in CommonDispatcher#dispatchCollectData , when `job.isSd()` is false,the condition `metricsSet == null` won't be checked before accessing `metricsSet`.

https://github.com/apache/hertzbeat/blob/63c187488f0b34701349d21931489f8cfefc964e/hertzbeat-collector/hertzbeat-collector-collector/src/main/java/org/apache/hertzbeat/collector/dispatch/CommonDispatcher.java#L301

This may lead to `NullPointerException`.
https://github.com/apache/hertzbeat/blob/63c187488f0b34701349d21931489f8cfefc964e/hertzbeat-collector/hertzbeat-collector-collector/src/main/java/org/apache/hertzbeat/collector/dispatch/CommonDispatcher.java#L305

## Checklist

## Add or update API

### DIFF
--- a/hertzbeat-collector/hertzbeat-collector-collector/src/main/java/org/apache/hertzbeat/collector/dispatch/CommonDispatcher.java
+++ b/hertzbeat-collector/hertzbeat-collector-collector/src/main/java/org/apache/hertzbeat/collector/dispatch/CommonDispatcher.java
@@ -298,7 +298,7 @@ public class CommonDispatcher implements MetricsTaskDispatch, CollectDataDispatc
                 }
             }
 
-            if (job.isSd() || metricsSet == null) {
+            if ( metricsSet == null || job.isSd() ) {
                 // The collection and execution of all metrics of this job are completed
                 // and the result listener is notified of the combination of all metrics data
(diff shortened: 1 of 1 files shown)


Calvin979:
@Rancho-7 Could you show the steps to reproduce this situation.
As I know, this problem can't be solve by switching their position in the coding
image
image
The results shown above are the same

Rancho-7:
Hi @Calvin979 You're right, this way can't fix the problem actually. Sorry for my misunderstanding. I tried reproducing it with the same way but couldn't - last time i saw this was by accident.

Calvin979:
> Hi @Calvin979 You're right, this way can't fix the problem actually. Sorry for my misunderstanding. I tried reproducing it with the same way but couldn't - last time i saw this was by accident.

Looking forward to your contribution!

Rancho-7:
Hi all, I'm planning to close this PR for now. I will reopen it once I figure out the root cause. Thanks!

Issue #3425, opened 2025-06-04
[BUG] <title>监控prometheus不能触发告警，版本：1.7.1


xudemo:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

可以正常监听到prometheus的数据，设置告警规则，告警中心一直收不到通知。

![Image](https://github.com/user-attachments/assets/551ec4a8-a9b9-4030-9fe8-a182af316b81)

![Image](https://github.com/user-attachments/assets/cce7e63a-115f-4acd-a81d-5bb6f30af309)

### Expected Behavior

根据告警规则，能触发告警通知

### Environment

```markdown
HertzBeat version(s):1.7
```

pwallk:
When adding Prometheus monitoring, the app will change to `_prometheus_` + monitorName, so it cannot match the rules

![Image](https://github.com/user-attachments/assets/0d708cba-3c3b-4c5b-a839-d5fc2398d430)

xudemo:
> When adding Prometheus monitoring, the app will change to `_prometheus_` + monitorName, so it cannot match the rules
> 
> ![Image](https://github.com/user-attachments/assets/0d708cba-3c3b-4c5b-a839-d5fc2398d430)

哥们，我需要如何修改，有解决方案吗？

pwallk:
> > When adding Prometheus monitoring, the app will change to `_prometheus_` + monitorName, so it cannot match the rules
> > ![Image](https://github.com/user-attachments/assets/0d708cba-3c3b-4c5b-a839-d5fc2398d430)
> 
> 哥们，我需要如何修改，有解决方案吗？

If you use `greptime` or `Victoria metrics`, you can temporarily use [Periodic Threshold](https://hertzbeat.apache.org/docs/help/alert_threshold/#scheduled-threshold)

xudemo:
> > > When adding Prometheus monitoring, the app will change to `_prometheus_` + monitorName, so it cannot match the rules
> > > ![Image](https://github.com/user-attachments/assets/0d708cba-3c3b-4c5b-a839-d5fc2398d430)
> > 
> > 
> > 哥们，我需要如何修改，有解决方案吗？
> 
> If you use `greptime` or `Victoria metrics`, you can temporarily use [Periodic Threshold](https://hertzbeat.apache.org/docs/help/alert_threshold/#scheduled-threshold)

可以给一个PromQL示例吗，我测试过多种写法都无法监听

[...]

PR #3434, opened 2025-06-05
[bugfix] Fix the issue where Prometheus RealTime Threshold is not eff…

Duansg:
## What's changed?

Please refer to:(#3425)

1、Because the monitor name was converted in `org.apache.hertzbeat.manager.service.impl.MonitorServiceImpl#addMonitor`
```
if (CommonConstants.PROMETHEUS.equals(monitor.getApp())) {
	appDefine.setApp(CommonConstants.PROMETHEUS_APP_PREFIX + monitor.getName());
}
```

2、And in `org.apache.hertzbeat.manager.service.impl.MonitorServiceImpl#addMonitor`, the label key: instance name is populated into the Job metadata.
```
Map<String, String> metadata = Map.of(CommonConstants.LABEL_INSTANCE_NAME, monitor.getName(), CommonConstants.LABEL_INSTANCE_HOST, monitor.getHost());
appDefine.setMetadata(metadata);
```

3、After the prometheus data has been collected, an alert needs to be sent to calculate whether the threshold is matched or not
```
// Send：
org.apache.hertzbeat.collector.dispatch.CommonDispatcher#dispatchCollectData(org.apache.hertzbeat.common.timer.Timeout, org.apache.hertzbeat.common.entity.job.Metrics, java.util.List<org.apache.hertzbeat.common.entity.message.CollectRep.MetricsData>)

// Calculate：
org.apache.hertzbeat.alert.calculate.RealTimeAlertCalculator#calculate
->
String app = metricsData.getApp();
// Filter thresholds by app, metrics, labels and instance  
thresholds = filterThresholdsByAppAndMetrics(thresholds, app, metrics, labels, instance, priority);  
if (thresholds.isEmpty()) {  
    return;  
}
```

4、The problem lies in the conversion in step 1, and I considered changing the page to backfill the monitor name, but then realized that it doesn't make sense.

I think a reasonable approach is to use (PROMETHEUS_APP_PREFIX + job metadata instanceName) to check if the current monitor is compatible. If it is, directly replace it with prometheus; if not, keep it as is

```
String app = metricsData.getApp();  
if ((CommonConstants.PROMETHEUS_APP_PREFIX + instanceName).equals(metricsData.getApp())) {  
    app = CommonConstants.PROMETHEUS;  
}
```

5、This is done for several reasons:
1. For multiple monitor jobs, each is isolated by a separate MetricsCollect.
2. Therefore, the data collected by each MetricsCollect is sent to alert using the instance name of the current job
3. For the calculation logic, it also determines whether a single job matches the calculation. Even when the threshold rule is associated with multiple monitoring configurations, since instanceId is treated as "or," it only requires the instance name (app) to remain consistent.

I haven't been familiar with the project for long; these are some of my thoughts and changes. Please let me know if anything is incorrect.
iShot_2025-06-05_22 45 30
iShot_2025-06-05_22 45 05
iShot_2025-06-05_22 44 31

## Checklist

## Add or update API

### DIFF
--- a/hertzbeat-alerter/src/main/java/org/apache/hertzbeat/alert/calculate/RealTimeAlertCalculator.java
+++ b/hertzbeat-alerter/src/main/java/org/apache/hertzbeat/alert/calculate/RealTimeAlertCalculator.java
@@ -139,6 +139,9 @@ public class RealTimeAlertCalculator {
         String instanceHost = metricsData.getInstanceHost();
         String app = metricsData.getApp();
         String metrics = metricsData.getMetrics();
+        if ((CommonConstants.PROMETHEUS_APP_PREFIX + instanceName).equals(metricsData.getApp())) {
+            app = CommonConstants.PROMETHEUS;
+        }
         int priority = metricsData.getPriority();
         int code = metricsData.getCode().getNumber();
         Map<String, String> labels = metricsData.getLabels();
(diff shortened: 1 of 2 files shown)


pwallk (approved):
lgtm

PR #3483, opened 2025-06-20
[fix]bugfix:Prevent Long.parseLong() error when search param is a float

starryCoder:
When the search parameter contains a float value (e.g., "9.111"), CommonUtil.isNumeric() returns true, but subsequent Long.parseLong() throws a NumberFormatException since floats cannot be converted to Long. This breaks the query flow.
## What's changed?
​​1.Switched from `Long.parseLong` to `Longs.tryParse` to safely parse ID strings, avoiding failures on large numbers or float.

## Checklist

## Add or update API


hertzbeat-common/src/main/java/org/apache/hertzbeat/common/util/CommonUtil.java
```diff
@@ -96,6 +98,18 @@ public static boolean isNumeric(String str) {
...
+     * whether the string is Long
+     * @param str string
+     * @return boolean
+     */
+    public static boolean isLong(String str) {
+        if (StringUtils.isBlank(str)) {
+            return false;
+        }
+        return LONG_PATTERN.matcher(str).matches();
+    }
```
tomsun28: hi can you help add some unit test cases?

starryCoder: Of course!

starryCoder: > hi can you help add some unit test cases?

Hi, I've added unit tests for isLong, and I noticed that the mock behavior covers the exception (monitorDao.findAll() directly returns mockPage without actually executing the query). Then I ensured test coverage by triggering toPredicate() with thenAnswer().

hertzbeat-manager/src/main/java/org/apache/hertzbeat/manager/service/impl/MonitorServiceImpl.java
```diff
@@ -608,7 +608,7 @@ public Page<Monitor> getMonitors(List<Long> monitorIds, String app, String searc
             if (StringUtils.hasText(search)) {
                 Predicate predicateHost = criteriaBuilder.like(root.get("host"), "%" + search + "%");
                 Predicate predicateName = criteriaBuilder.like(criteriaBuilder.lower(root.get("name")), "%" + search.toLowerCase() + "%");
-                if (CommonUtil.isNumeric(search)){
+                if (CommonUtil.isLong(search)){
                     Predicate predicateId = criteriaBuilder.equal(root.get("id"), Long.parseLong(search));
```
pjfanning: Long.parseLong can still fail if the number has lots of digits, eg 20 or more.
9,223,372,036,854,775,807 is the max long in Java.

starryCoder: Hi，I have switched from `Long.parseLong` to `Longs.tryParse` to safely parse ID strings, avoiding failures on large numbers or float.

Issue #3496, opened 2025-06-23
[Question] PromQL comparison operator implementation is inconsistent with the official Prometheus definition


bigcyy:
### Question

While recently reviewing the official Prometheus documentation, I noticed the behavior of PromQL comparison operators (>, <, ==, >=, <=, !=) and found two main inconsistencies when comparing it with the implementation in Hertzbeat.

1. Comparison Logic for Instant Vector vs. Scalar

    According to the [official Prometheus documentation](https://prometheus.io/docs/prometheus/latest/querying/operators/), when an instant vector is compared to a scalar, the operation acts as a filter. Specifically, PromQL iterates through each time series in the vector, performs the comparison, and any time series for which the expression is true is kept in the output. Series for which the result is false are removed.

    However, I noticed that the behavior demonstrated in the calculate1 test case in hertzbeat.warehouse.service.DataSourceServiceTest does not align with this standard.

    ```java
    void calculate1() {
        List<Map<String, Object>> prometheusData = List.of(
                new HashMap<>(Map.of("__value__", 100.0, "timestamp", 1343554, "instance", "node1")),
                new HashMap<>(Map.of("__value__", 200.0, "timestamp", 1343555, "instance", "node2"))
        );
    
        QueryExecutor mockExecutor = Mockito.mock(QueryExecutor.class);
        Mockito.when(mockExecutor.support("promql")).thenReturn(true);
        Mockito.when(mockExecutor.execute(Mockito.anyString())).thenReturn(prometheusData);
        dataSourceService.setExecutors(List.of(mockExecutor));
    
        List<Map<String, Object>> result = dataSourceService.calculate("promql", "node_cpu_seconds_total > 150");
        assertEquals(2, result.size());
        assertNull(result.get(0).get("__value__"));
        assertEquals(200.0, result.get(1).get("__value__"));
    }
    ```
    Analysis:

    Input: An instant vector containing two time series with values 100.0 and 200.0.
    Query: node_cpu_seconds_total > 150.
    Expected PromQL Behavior: Only the series with the value 200.0 satisfies the condition. Therefore, the result should be a new vector containing only that single series, and its size should be 1.
    Behavior in Hertzbeat Test: The test asserts that result.size() is 2. Instead of filtering out the non-matching series (__value__ of 100.0), it keeps the series in the result set but sets its __value__ to null.
    Core Question:
[...]

PR #3505, opened 2025-06-24
[feature] Support for `Periodic Threshold` preview.

Duansg:
## What's changed?

Support for `Periodic Threshold` preview.

For details:
1. Added `Threshold` -> `Periodic Threshold` preview, exception content support for internationalization
2. Add anltr4 `Look Ahead` method to determine if parsing is finished.
3. Add new test cases/fix test cases and go through the history of test cases successfully.

## Checklist

## Add or update API

iShot_2025-06-24_20 15 55
iShot_2025-06-24_20 21 25

1、andOp
and

2、compareOp

3、orOp
OR

### DIFF
--- a/hertzbeat-alerter/src/main/java/org/apache/hertzbeat/alert/service/impl/AlertDefineServiceImpl.java
+++ b/hertzbeat-alerter/src/main/java/org/apache/hertzbeat/alert/service/impl/AlertDefineServiceImpl.java
@@ -17,7 +17,6 @@
 
 package org.apache.hertzbeat.alert.service.impl;
 
-import static org.apache.hertzbeat.common.constants.CommonConstants.ALERT_THRESHOLD_TYPE_REALTIME;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -29,6 +28,7 @@ import org.apache.hertzbeat.alert.calculate.PeriodicAlertRuleScheduler;
 import org.apache.hertzbeat.alert.dao.AlertDefineDao;
 import org.apache.hertzbeat.alert.service.AlertDefineImExportService;
 import org.apache.hertzbeat.alert.service.AlertDefineService;
+import org.apache.hertzbeat.alert.service.DataSourceService;
 import org.apache.hertzbeat.common.cache.CacheFactory;
 import org.apache.hertzbeat.common.constants.ExportFileConstants;
 import org.apache.hertzbeat.common.constants.SignConstants;
@@ -58,6 +58,9 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
+import static org.apache.hertzbeat.common.constants.CommonConstants.ALERT_THRESHOLD_TYPE_PERIODIC;
+import static org.apache.hertzbeat.common.constants.CommonConstants.ALERT_THRESHOLD_TYPE_REALTIME;
+
 /**
  * Alarm definition management interface implementation
  */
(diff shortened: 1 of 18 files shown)


MasamiYui:
looks great!

Duansg:
> looks great!

hh, review it for me if you have time. :wink:

hertzbeat-alerter/src/main/resources/alerter_zh_CN.properties
```diff
@@ -32,3 +32,4 @@ alerter.notify.console = 登入控制台
 alerter.priority.0 = 紧急告警
 alerter.priority.1 = 严重告警
 alerter.priority.2 = 警告告警
+alerter.calculate.parse.error = 表达式未完全解析，可能存在语法错误或输入不完整
```
MasamiYui, line 35: Make supplementary modifications in alerter_zh_TW.properties

hertzbeat-alerter/src/main/java/org/apache/hertzbeat/alert/service/impl/AlertDefineServiceImpl.java
```diff
@@ -218,4 +224,18 @@ public List<AlertDefine> getRealTimeAlertDefines() {
...
         return alertDefines;
     }
+
+    @Override
+    public List<Map<String, Object>> getDefinePreview(String datasource, String type, String expr) {
+        if (!StringUtils.hasText(expr) || !StringUtils.hasText(datasource) || !StringUtils.hasText(type)) {
+            return Collections.emptyList();
+        }
+        switch (type) {
+            case ALERT_THRESHOLD_TYPE_PERIODIC:
```
MasamiYui: Enum suggest using CommonConstants.xxx

MasamiYui:
@tomsun28 @bigcyy What are your thoughts on this preview feature? Personally, From a UI perspective, as an common user, I might not fully grasp the meaning conveyed here, such as inner elements like `__value__`

hertzbeat-alerter/src/main/java/org/apache/hertzbeat/alert/service/impl/AlertDefineServiceImpl.java
Duansg: Adjusted and changed in two other places in this.

Duansg:
> @tomsun28 @bigcyy What are your thoughts on this preview feature? Personally, From a UI perspective, as an common user, I might not fully grasp the meaning conveyed here, such as inner elements like `__value__`

I don't think there is anything to worry about for the following reasons.

1. If `__value__` doesn't make sense, we can change it to `value`.
2. If `value` is not understandable, then I would suggest the user to read the official documentation.

bigcyy:
Looks good, but would it be better to display it as shown in the picture? My reasons are as follows:
Due to the 'and' set operations, if a structured table is used, the table header might be very long. Perhaps JSON would be more suitable.

IMG_4561

Duansg:
> Looks good, but would it be better to display it as shown in the picture? My reasons are as follows: Due to the 'and' set operations, if a structured table is used, the table header might be very long. Perhaps JSON would be more suitable.
> 
> IMG_4561

That's good advice, I'll have time to deal with it tomorrow.

MasamiYui:
@Duansg How about adding a marker to the corresponding row if it can trigger an alert?

Duansg:
> How about adding a marker to the corresponding row if it can trigger an alert?

@MasamiYui Sorry, I didn't quite get your point, could you expand on it. :)

MasamiYui:
Like this:

image

`__VALUE__ == null` -> No

`__VALUE__  != null` -> YES

Duansg:
> Like this:
> 
> image
> `__VALUE__ == null` -> No
> 
> `__VALUE__ != null` -> YES

@MasamiYui Emmm, this is indeed in line with the current design, but I think we should not add this description for the time being, for the following reasons

1. After our last discussion, I still firmly believe that we must ensure the integrity and scalability of the parser and ensure consistency with the `promql` semantics, that is, abandon the judgment of `value == null` and replace it with `metrics == null`. Otherwise, adding hertzbeat's unique judgment to the parser will make the parser very difficult to maintain and the complexity of achieving semantic correctness.
2. If the above functional effects are achieved, the result of the `preview` function will be what you see is what you get, just like the search in the prometheus ui, so there is no need for a special mark to tell the user whether this item is hit, otherwise if a large amount of missed data is a terrible rendering function point.

Therefore, I think that it is enough to achieve the effect of `basic preview` at present. After we complete the relevant logic of the parser in the future, we will come back to see if there are other optimization points for this content. What do you think?

Duansg:
> Looks good, but would it be better to display it as shown in the picture? My reasons are as follows: Due to the 'and' set operations, if a structured table is used, the table header might be very long. Perhaps JSON would be more suitable.
> 
> IMG_4561

@MasamiYui @bigcyy 

Sorry friends, I'm busy with work recently. I will continue to improve this PR tomorrow. If you are all free at that time, please review it for me. Thanks. 😅

bigcyy:
> 1. I still firmly believe that we must ensure the integrity and scalability of the parser and ensure consistency with the `promql` semantics

I agree, see #3496 .

Duansg:
1、orOp
OR
2、andOp
and
3、comOp
com
3、error display
error
5、unlessOp
unless
4、metrics data is empty
iShot_2025-06-28_17 03 27

PR #3574, opened 2025-07-13
[refactor] Refactor the comparison logic

Duansg:
## What's changed?

Refactor the comparison logic in `org.apache.hertzbeat.alert.expr.AlertExpressionEvalVisitor#visitComparisonExpr`

For details:
1. Optimized the logic of historical `comparison binary operators`.
2. Improve the semantics of promql, add the `comparison binary operators` function, and add `scalar and scalar`, `scalar and vector`, and `vector and vector`.
3. Antlr4 added the `bool` modifier.
iShot_2025-07-13_23 24 45

4. Compatible with current alarm logic processing.
5. Complete/fix test cases and add new test cases.

## Checklist

## Add or update API

// vector and vector
iShot_2025-07-13_21 48 51
iShot_2025-07-13_21 49 16
iShot_2025-07-13_21 53 03

// vector and scalar
iShot_2025-07-13_21 47 27
iShot_2025-07-13_21 47 49
iShot_2025-07-13_21 48 15

// scalar and vector
iShot_2025-07-13_21 45 26
iShot_2025-07-13_21 46 00
iShot_2025-07-13_21 46 27

// scalar and scalar
iShot_2025-07-13_21 43 37
iShot_2025-07-13_21 43 51

### DIFF
--- a/hertzbeat-manager/src/main/java/org/apache/hertzbeat/manager/controller/GeneralConfigController.java
+++ b/hertzbeat-manager/src/main/java/org/apache/hertzbeat/manager/controller/GeneralConfigController.java
@@ -17,7 +17,6 @@
 
 package org.apache.hertzbeat.manager.controller;
 
-import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.Parameter;
 import io.swagger.v3.oas.annotations.tags.Tag;
@@ -25,6 +24,7 @@ import jakarta.annotation.Resource;
 import jakarta.validation.constraints.NotNull;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.hertzbeat.common.entity.dto.Message;
+import org.apache.hertzbeat.common.util.CommonUtil;
 import org.apache.hertzbeat.common.util.ResponseUtil;
 import org.apache.hertzbeat.manager.pojo.dto.TemplateConfig;
 import org.apache.hertzbeat.manager.service.ConfigService;
@@ -37,6 +37,19 @@ import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
 
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.time.format.TextStyle;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
+
 /**
  * Alert sender Configuration API
  */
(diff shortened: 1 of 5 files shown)


MasamiYui:
Thank you for your PR. 

There is an example I would like to discuss with you. According to the current parsing rules, `http_requests_time > avg(rate(http_requests_time[5m]))` will be parsed as `vector > vector`, but in reality, it should be `vector > scalar`. 

At present, only LiteralExpr(number) is regarded as a scalar. I think more judgment conditions may be needed to determine whether something is a scalar, such as whether there are labels in the metric query results.

Duansg:
> avg(rate(http_requests_time[5m]))



> Thank you for your PR.
> 
> There is an example I would like to discuss with you. According to the current parsing rules, `http_requests_time > avg(rate(http_requests_time[5m]))` will be parsed as `vector > vector`, but in reality, it should be `vector > scalar`.
> 
> At present, only LiteralExpr(number) is regarded as a scalar. I think more judgment conditions may be needed to determine whether something is a scalar, such as whether there are labels in the metric query results.

Hi, @MasamiYui  thank you for your review. There are a few points regarding your question.

1. As long as your query is of type "instant vector" (even if there is only one value), the result will be a vector. You can verify this in the Prometheus UI and HTTP API.
2. Aggregate functions such as `avg`, which you mentioned in your example, are not scalars in the true sense of the word. I consider them to be unlabeled scalars, and I believe that the expression you provided will not return a result.
3. The normal approach would be to use the `scalar` function to return the sample value as a scalar, but the current `PromqlQueryExecutor` does not have the ability to parse `resultType`. I am currently working on refactoring this.


image

MasamiYui:
> > avg(rate(http_requests_time[5m]))
> 
> > Thank you for your PR.
> > There is an example I would like to discuss with you. According to the current parsing rules, `http_requests_time > avg(rate(http_requests_time[5m]))` will be parsed as `vector > vector`, but in reality, it should be `vector > scalar`.
> > At present, only LiteralExpr(number) is regarded as a scalar. I think more judgment conditions may be needed to determine whether something is a scalar, such as whether there are labels in the metric query results.
> 
> Hi, @MasamiYui thank you for your review. There are a few points regarding your question.
> 
> 1. As long as your query is of type "instant vector" (even if there is only one value), the result will be a vector. You can verify this in the Prometheus UI and HTTP API.
> 2. Aggregate functions such as `avg`, which you mentioned in your example, are not scalars in the true sense of the word. I consider them to be unlabeled scalars, and I believe that the expression you provided will not return a result.
> 3. The normal approach would be to use the `scalar` function to return the sample value as a scalar, but the current `PromqlQueryExecutor` does not have the ability to parse `resultType`. I am currently working on refactoring this.
> 
> image

👍  I got it!

Duansg:
@MasamiYui Oh, right, as of this PR, the semantic correctness should all be handled. Next, we should discuss the plans we mentioned earlier. When is the next meeting? We can discuss that.

I remember @tomsun28  saying that the meeting time was posted on the public channel. How can I subscribe to this information?

MasamiYui:
> @MasamiYui Oh, right, as of this PR, the semantic correctness should all be handled. Next, we should discuss the plans we mentioned earlier. When is the next meeting? We can discuss that.
> 
> I remember @tomsun28 saying that the meeting time was posted on the public channel. How can I subscribe to this information?

@Duansg  please ref to : [mailing_lists](https://hertzbeat.apache.org/docs/community/mailing_lists)

PR #3595, opened 2025-07-20
[improve] VM write request sets up gzip compression and adds saveData unit tests

Cyanty:
## What's changed?

- Refer to the victoriametrics doc to set gzip compression for write requests.
```sh
# Import gzipped data to <destination-victoriametrics>:
curl -X POST -H 'Content-Encoding: gzip' http://destination-victoriametrics:8428/api/v1/import -T exported_data.jsonl.gz
```

- Add unit tests for inserting historical metric data, test the execution process of insertion for buffer size and flush interval.

- Update the configuration parameters of `application-test.yml`.

## Checklist

## Add or update API

### DIFF
--- a/hertzbeat-manager/src/main/resources/define/app-rabbitmq.yml
+++ b/hertzbeat-manager/src/main/resources/define/app-rabbitmq.yml
@@ -21,11 +21,13 @@ app: rabbitmq
 name:
   zh-CN: RabbitMQ消息系统
   en-US: RabbitMQ Message
+  ja-JP: RabbitMQメッセージシステム
 # The description and help of this monitoring type
 help:
   zh-CN: HertzBeat 对 RabbitMQ 消息中间件的运行状态，节点，队列等相关指标（overview、object_totals、nodes、queues）进行监测。<br><span class='help_module_span'>⚠️注意：HertzBeat 使用 RabbitMQ Management 的 Rest Api 对 RabbitMQ 进行指标数据采集，因此您的 RabbitMQ 环境需要开启 Management 插件，<a class='help_module_content' href='https://hertzbeat.apache.org/zh-cn/docs/help/rabbitmq#%E7%9B%91%E6%8E%A7%E5%89%8D%E6%93%8D%E4%BD%9C'>点击查看开启步骤</a>。</span>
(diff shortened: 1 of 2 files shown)


hertzbeat-warehouse/src/main/java/org/apache/hertzbeat/warehouse/store/history/tsdb/vm/VictoriaMetricsDataStorage.java
```diff
@@ -625,11 +627,24 @@ private void doSaveData(List<VictoriaMetricsContent> contentList) {
...
             }
-            HttpEntity<String> httpEntity = new HttpEntity<>(stringBuilder.toString(), headers);
+            String payload = stringBuilder.toString();
+            // compress the payload using gzip
+            byte[] compressedPayload;
+            try (ByteArrayOutputStream bos = new ByteArrayOutputStream(payload.length());
+                 GZIPOutputStream gzip = new GZIPOutputStream(bos)) {
+                gzip.write(payload.getBytes(StandardCharsets.UTF_8));
+                // gzip stream must be turned off to complete the compression
+                gzip.close();
```
Calvin979: Why need to call `close` here, `GZIPOutputStream` is already defined within `try` keyword.
I've found a method `finish` in `GZIPOutputStream`, is it the method you wanna execute here?

Cyanty: Yes, "finish" is more correct. I will replace "close" with it. Thanks！

MasamiYui (commented):
Hi, thanks for your PR. Here's a small suggestion: How about compression be made a configurable option (such as content-encoding)? This is because it may not be suitable for some scenarios, like small text.

Cyanty:
> Hi, thanks for your PR. Here's a small suggestion: How about compression be made a configurable option (such as content-encoding)? This is because it may not be suitable for some scenarios, like small text.

No problem. I will re-commit after adding the compression parameter configuration items.

Cyanty:
> hi, the following also needs to be modified
script/docker-compose/hertzbeat-mysql-victoria-metrics/conf/application.yml
script/docker-compose/hertzbeat-postgresql-victoria-metrics/conf/application.yml

ok, 🤞I forgot that there are also configuration items for this part. thank you for pointing that out.

PR #3604, opened 2025-07-23
[feat] Auto-generate AES key if not configured

bigcyy:
## What's changed?

This pull request enhances the AES encryption module to improve the initial setup process and default security posture.

1. If the application starts and cannot find a pre-existing AES key in the database and application.yml, it will now automatically generate a strong, unique key and persist it for future use. 

2. When decrypting data, the system will first try the new (auto-generated or user-configured) key. If decryption fails (e.g., on data encrypted with an older key), it will automatically fall back and attempt to decrypt using the legacy default key. This guarantees that old data remains accessible after the upgrade.

## Checklist

## Add or update API

### DIFF
--- a/hertzbeat-manager/src/main/java/org/apache/hertzbeat/manager/config/ConfigInitializer.java
+++ b/hertzbeat-manager/src/main/java/org/apache/hertzbeat/manager/config/ConfigInitializer.java
@@ -28,6 +28,7 @@ import lombok.SneakyThrows;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hertzbeat.common.constants.CommonConstants;
 import org.apache.hertzbeat.common.entity.manager.GeneralConfig;
+import org.apache.hertzbeat.common.util.AesUtil;
 import org.apache.hertzbeat.common.util.TimeZoneUtil;
 import org.apache.hertzbeat.base.dao.GeneralConfigDao;
 import org.apache.hertzbeat.manager.pojo.dto.MuteConfig;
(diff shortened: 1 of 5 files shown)


hertzbeat-manager/src/main/java/org/apache/hertzbeat/manager/config/ConfigInitializer.java
```diff
@@ -111,29 +115,38 @@ public void initConfig() {
...
+            // use the random aes secret
+            SystemSecret systemSecret = systemSecretService.getConfig();
+            if (systemSecret == null || StringUtils.isBlank(systemSecret.getAesSecret())) {
+                currentAesSecret = randomizeSecret(AesUtil.ENCODE_RULES);
+                builder.aesSecret(currentAesSecret);
+            } else {
+                currentAesSecret = systemSecret.getAesSecret();
+            }
+        }
+        AesUtil.setDefaultSecretKey(currentAesSecret);
```
tomsun28, line 145: Hi, @bigcyy @zqr10159 @yuluo-yx   I found that we overlooked an important point. The AesUtil will be use in selfhosted `collector`, which need use the same AesSecret with the `hertzbeat-manager`, else the `hertzbeat-collector` can not `AesUtil.decode` the encrypted string.

tomsun28: How about this plan?
The selfhosted `hertzbeat-collector` will send a register message to `hertzbeat-manager` when first startup, the `hertzbeat-manager` can response with the message that contains same serect, so then the collector can use it to decode the encrypted string.

Issue #3623, opened 2025-08-01
[BUG] mysql使用jdbc url进行连接时，配置参数失效


huisunan:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

jdbc:mysql://localhost/test?allowPublicKeyRetrieval=true --->jdbc:mysql://localhost/test?allowpublickeyretrieval=true

#3428 

@TJxiaobao 

### Expected Behavior

校验时不改变原有参数

### Environment

```markdown
HertzBeat version(s): v1.7.2
```

Duansg:
:thumbsup: Hi @huisunan ,  thank you for raising this issue. I will reproduce and fix it.

PR #3625, opened 2025-08-02
[fix] Fixed issue where jdbc url was replaced

Duansg:
## What's changed?

Please refer to: [#3623](https://github.com/apache/hertzbeat/issues/3623)

The reason is that during JDBC URL security-related processing, the returned URL is replaced by the system. We believe that the case-sensitive recommendations in [connector-j-reference-jdbc-url-format](https://dev.mysql.com/doc/connectors/en/connector-j-reference-jdbc-url-format.html) should be followed to enable security-related processing without replacing user-entered URLs.

For details:
1. Fixed an issue where the Jdbc URL was replaced and ensured security-related processing.
2. Added a test case to verify whether the Jdbc URL is consistent with the user input after processing is complete.
3. Added test cases for security-related processing.

## Checklist

## Add or update API


TJxiaobao (approved):
👍👍👍

Issue #3630, opened 2025-08-04
[Task] Monitor template definition JEXL keywords check and repair


Duansg:
### Description

Hi all , currently, JEXL contains some reserved keywords, but there are still some keyword configurations in the Hertzbeat monitoring template. If used improperly, this may cause parsing errors. To resolve potential parsing issues that may arise in the future, it is necessary to check and fix this potential issue.

For more details, please refer to PR: [#3629](https://github.com/apache/hertzbeat/pull/3629)
-- 

What needs to be done:
--- 
1. Check and reproduce keyword anomalies.
2. Fix keyword issues
3. Ensure normal process availability.
4. Please provide test cases if necessary.

Reference script: 
--- 
`org.apache.hertzbeat.manager.script.YamlCheckScript`

Reference keywords: 
--- 
[keywords](https://github.com/apache/commons-jexl/blob/master/src/main/java/org/apache/commons/jexl3/parser/Parser.jjt)

### Task List

- [x] app-postgresql.yml
- [x] app-spring_gateway.yml
- [x] app-windows.yml
- [x] app-hertzbeat.yml
- [ ] app-vastbase.yml
- [x] app-hertzbeat_token.yml
- [x] app-kingbase.yml
- [x] app-springboot2.yml @delei 
- [x] app-springboot3.yml @Duansg 
- [ ] app-greenplum.yml

delei:
Hi, @Duansg 
Are the yml files in these task lists confirmed to contain keywords that need to be processed?
I'm willing to try to fix the `app-springboot2.yml` template.

Duansg:
> Hi, [@Duansg](https://github.com/Duansg) Are the yml files in these task lists confirmed to contain keywords that need to be processed? I'm willing to try to fix the `app-springboot2.yml` template.

:thumbsup: Welcome, these are all files containing keywords, please confirm them.

delei:
Hi, @Duansg 
1.7.3 Version Milestone (Due by 23 August 2025)

I noticed that there are still many templates that have not been fixed. Is it necessary to submit a PR for each file separately, or should we run the `YamlCheckScript#checkYaml()` test to ensure that template can be saved successfully?

Duansg:
> Hi, [@Duansg](https://github.com/Duansg) 1.7.3 Version Milestone (Due by 23 August 2025)
> 
[...]

Issue #3626, opened 2025-08-02
[BUG] Threshold configuration expression error


Duansg:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When I tried to save the Threshold configuration for Spring Boot 3, the system prompted an expression error. I tried Spring Boot 2 and encountered a similar error.

### Expected Behavior

Threshold configuration can be selected normally and saved successfully

### Steps To Reproduce

<img width="1496" height="608" alt="Image" src="https://github.com/user-attachments/assets/d12216b4-ceff-403e-9bf2-8564edd749f7" />

### Environment

```markdown
HertzBeat version(s): 1.7.2
```

Duansg:
I have identified the cause of this error and will fix it today

PR #3634, opened 2025-08-05
fix: quote SQL strings containing JEXL reserved keywords in monitoring templates

SarthakMeshram:
This PR fixes potential parsing issues caused by JEXL reserved keywords in SQL strings in monitoring templates.
Changes:
- Added quotes around SQL strings containing JEXL reserved keywords.
- Updated 11 monitoring template files.

Related to issue: #3630

### DIFF
--- a/hertzbeat-manager/src/main/resources/define/app-dm.yml
+++ b/hertzbeat-manager/src/main/resources/define/app-dm.yml
@@ -161,7 +161,8 @@ metrics:
       # SQL Query Method：oneRow, multiRow, columns
       queryType: columns
       # script sql
-      sql: select PARA_NAME, PARA_VALUE from SYS."V$DM_INI" where PARA_NAME = 'MAX_SESSIONS'or PARA_NAME = 'CTL_PATH' or PARA_NAME = 'PORT_NUM';
+      sql: "'select PARA_NAME, PARA_VALUE from SYS.\"V$DM_INI\" where PARA_NAME = ''MAX_SESSIONS'' or PARA_NAME = ''CTL_PATH'' or PARA_NAME = ''PORT_NUM'';'"
+
       # JDBC url
       url: ^_^url^_^
 
--- a/hertzbeat-manager/src/main/resources/define/app-greenplum.yml
+++ b/hertzbeat-manager/src/main/resources/define/app-greenplum.yml
@@ -162,7 +162,7 @@ metrics:
       # SQL Query Method：oneRow, multiRow, columns
       queryType: columns
       # sql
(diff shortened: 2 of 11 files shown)


Duansg:
Hi, @SarthakMeshram first of all, thank you for submitting your PR. However, I think there may be a misunderstanding.

This task is an extension of issue [#3626], because when configuring thresholds, if the `field` contains JEXL keyword information, the final generated expression will fail due to JEXL compilation.

However, after reviewing your modifications, I noticed that `jdbc.sql` did not undergo JEXL compilation and calculation. Is this PR associated with an error, or did you encounter another issue?

We look forward to hearing from you.

SarthakMeshram:
Hi @Duansg,
Thank you for the feedback and for pointing this out. I realize now that I misunderstood the task — I had focused on SQL queries containing keywords like or/and, but I see from your explanation that the real issue is with field or metric names in the monitoring templates that match JEXL reserved keywords, which can break threshold expressions during JEXL compilation.

I’d like to correct my approach by running YamlCheckScript to identify these keywords in the listed YAML files and fixing them accordingly. Could you please confirm if the preferred fix is to rename the fields or quote them in the YAML? Once I know the preferred approach, I’ll update my PR.

Thanks again for the clarification.

Duansg:
> Hi @Duansg, Thank you for the feedback and for pointing this out. I realize now that I misunderstood the task — I had focused on SQL queries containing keywords like or/and, but I see from your explanation that the real issue is with field or metric names in the monitoring templates that match JEXL reserved keywords, which can break threshold expressions during JEXL compilation.
> 
> I’d like to correct my approach by running YamlCheckScript to identify these keywords in the listed YAML files and fixing them accordingly. Could you please confirm if the preferred fix is to rename the fields or quote them in the YAML? Once I know the preferred approach, I’ll update my PR.
> 
> Thanks again for the clarification.

@SarthakMeshram Thanks for your reply.

Actually, there are other ways to handle this, such as using `obj["keyword"]` or adding backticks, which will make it a "normal variable" rather than a keyword.

However, these changes are quite extensive, involving expression generation methods and other aspects, so they may take a long time to implement and require extensive regression testing. They may also introduce new issues.

Therefore, renaming and performing preliminary verification is currently the most appropriate approach given the current circumstances.

delei:
> Therefore, renaming and performing preliminary verification is currently the most appropriate approach given the current circumstances.

Hi, @SarthakMeshram I think it should be kept to minimal changes as well . I have submitted several separate PRs to handle and validate each separately, please take a look at this PR #3685 , and see if it is helpful to you.

cc @Duansg 
If you feel that this is not appropriate, please let us know at any time.

SarthakMeshram:
Thank you for the review and guidance. Since the related issue #3630 has already been resolved in #3708, I’ll go ahead and close this PR. I appreciate the feedback and will look for another issue to contribute to.

PR #3641, opened 2025-08-07
[feature] Support HertzBeat self monitor

zqr10159:
## What's changed?

prometheus metrics endpoint `http://[host]:1157/actuator/prometheus`
grafana dashboard json : `hertzbeat-manager/src/main/resources/grafana/hertzbeat-monitor-zh-CN.json`
image
8875ac7a74d7edad8f06947c01ab15f2
058d4cd48c272b85d5370bc5393373de
0463d10cf8297d702348d285bea6f6a8

## Checklist

- [ ]  I have read the [Contributing Guide](https://hertzbeat.apache.org/docs/community/code_style_and_quality_guide)
- [ ]  I have written the necessary doc or comment.
- [ ]  I have added the necessary unit tests and all cases have passed.

## Add or update API

- [ ] I have added the necessary [e2e tests](https://github.com/apache/hertzbeat/tree/master/e2e) and all cases have passed.

PR #3673, opened 2025-08-17
[OSPP] Apache HertzBeat Log Monitoring Capability

bigcyy:
## What's changed?

- [X] Integration with HertzBeat based on the OpenTelemetry log protocol.
- [X] Real-time log stream viewing.
- [X] Log storage.
- [X] Real-time log threshold alerting (JEXL matching, window aggregation, watermark mechanism).
- [X] Periodic log threshold alerting (SQL data query, JEXL matching).
- [X] Log management and statistics.
- [X] Unit tests.
- [x] E2E tests.
- [x] Update documentation.
- [ ] Detail refinement.

## Checklist

## Add or update API


hertzbeat-alerter/src/main/java/org/apache/hertzbeat/alert/AlerterWorkerPool.java
```diff
@@ -77,6 +79,21 @@ private void initNotifyExecutor() {
...
+        ThreadFactory threadFactory = new ThreadFactoryBuilder()
+                .setUncaughtExceptionHandler((thread, throwable) -> {
+                    log.error("Alerter logWorkerExecutor has uncaughtException.");
+                    log.error(throwable.getMessage(), throwable);
+                })
+                .setDaemon(true)
+                .setNameFormat("log-worker-%d")
+                .build();
+        logWorkerExecutor = new ThreadPoolExecutor(4, 10, 10, TimeUnit.SECONDS,
+                new LinkedBlockingQueue<>(),
```
tomsun28: hi, if the blocking queue is LinkedBlockingQueue, the thread size first will increase to 4, and then others tasks will put in the LinkedBlockingQueue, the thread size will never increase until the blocking queue is filled.

tomsun28:
Great!👍 👍 So big feature, we need more time to review.

web-app/src/app/routes/log/log-stream/log-stream.component.ts
```diff
@@ -0,0 +1,402 @@
...
+  // Add method to manually scroll to top
+  scrollToTop(): void {
+    this.userScrolled = false;
+    this.isNearBottom = true;
+    this.scheduleAutoScroll();
+  }
+
+  // Utility methods
+  getSeverityColor(severityNumber: number | undefined): string {
+    console.log('severityNumber', severityNumber);
```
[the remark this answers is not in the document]
tomsun28: +1 the console.log need be removed.

hertzbeat-alerter/src/main/java/org/apache/hertzbeat/alert/AlerterWorkerPool.java
bigcyy: Hi tom, thanks for your feedback! I've pushed a fix to address this.

web-app/src/assets/i18n/zh-TW.json
```diff
@@ -245,8 +245,8 @@
   "alert.setting.default.tip": "此告警阈值配置是否應用于全局所有此類型監控",
   "alert.setting.delete": "刪除阈值規則",
   "alert.setting.edit": "編輯阈值規則",
-  "alert.setting.edit.periodic": "编辑计划阈值",
-  "alert.setting.edit.realtime": "编辑实时阈值",
+  "alert.setting.edit.periodic": "編輯周期阈值",
```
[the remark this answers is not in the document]
bigcyy: down

web-app/src/app/routes/log/log-stream/log-stream.component.ts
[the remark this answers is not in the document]
bigcyy: down

hertzbeat-common/src/main/java/org/apache/hertzbeat/common/entity/alerter/AlertDefine.java
```diff
@@ -100,6 +100,11 @@ public class AlertDefine {
     @Size(max = 100)
     private String datasource;
 
+    @Schema(title = "Query Expression", example = "SELECT * FROM metrics WHERE value > 90")
+    @Size(max = 2048)
+    @Column(length = 2048)
+    private String queryExpr;
+
```
tomsun28: 👍 hi, overall it is very good. 
We need to discuss whether the queryExpr for log here is reasonable. In the previous design, the query expression and the judgment expre were together, such as the promql threshold rule like, expr: `(100 - (avg by(instance)(irate(node_cpu_seconds_total{mode="idle"}[5m])) * 100)) > 80`, you can see the `(100 - (avg by(instance)(irate(node_cpu_seconds_total{mode="idle"}[5m])) * 100))` is query expression and the `> 80` is the judgment expression.

hertzbeat-alerter/src/main/java/org/apache/hertzbeat/alert/reduce/AlarmCommonReduce.java
```diff
@@ -66,6 +68,22 @@ private void initWorkExecutor() {
...
+                    String fingerprint = generateAlertFingerprint(alert.getLabels());
+                    alert.setFingerprint(fingerprint);
+                }
+                // Process the group alert
+                alarmGroupReduce.processGroupAlert(groupLabels, alerts);
+            } catch (Exception e) {
+                log.error("Reduce alarm group failed: {}", e.getMessage());
+            }
+        });
+    }
```
tomsun28, line 86: here we are handling the group alerts separately with special processing, but the `reduce` are design for all alerts which from internal system and extern system.  I think we can keep it for now because no new database fields have been added.

hertzbeat-common/src/main/java/org/apache/hertzbeat/common/entity/alerter/AlertDefine.java
bigcyy: Hi Tom, thank you for your review. I believe that a two-stage design, separating the query and alerting phases, is of paramount importance. If we do not make this separation, our SQL query syntax will be tightly coupled with the alerting threshold syntax. This coupling would lead to insurmountable syntax conflicts when adapting to different SQL dialects, such as those used by Greptime and other databases.

Furthermore, from a user experience perspective, this two-stage model is more intuitive in UI design. It perfectly aligns with the user's natural mental model: first defining "What data do I need?" and then determining "What conditions should trigger an alert?" This approach is much easier to understand and operate than one that mixes all the logic together.

Finally, it is worth noting that this two-stage model is widely adopted in mature industry monitoring platforms, such as Alibaba Cloud's Log Service (SLS). This further validates the stability and maturity of a decoupled solution.

bigcyy: https://help.aliyun.com/zh/sls/configure-an-alert-monitoring-rule-in-log-service

Duansg: > Hi Tom, thank you for your review. I believe that a two-stage design, separating the query and alerting phases, is of paramount importance. If we do not make this separation, our SQL query syntax will be tightly coupled with the alerting threshold syntax. This coupling would lead to insurmountable syntax conflicts when adapting to different SQL dialects, such as those used by Greptime and other databases.
> 
> Furthermore, from a user experience perspective, this two-stage model is more intuitive in UI design. It perfectly aligns with the user's natural mental model: first defining "What data do I need?" and then determining "What conditions should trigger an alert?" This approach is much easier to understand and operate than one that mixes all the logic together.
> 
> Finally, it is worth noting that this two-stage model is widely adopted in mature industry monitoring platforms, such as Alibaba Cloud's Log Service (SLS). This further validates the stability and maturity of a decoupled solution.

I also have a question about this. If there is no prompt for the query field, do I still need to check the information displayed in the relevant log to correctly configure the threshold expression?

I use SLS more frequently in my daily work. In most cases, I only need phrase matching. When I need to format information or process other data, I use SQL to enhance my writing. The convenience lies in the fact that when I generate threshold expressions, I can quickly generate the required expressions through AI-powered query generation, historical queries, and prompts. Could we also adopt this approach?

Then I looked at the code you discussed for this position. As I understand it, you should first perform a `QueryExpr` query, then use `expr` to perform JEXL threshold matching. Wouldn't this be a bit complicated? I'm also concerned that there might be a performance bottleneck when executing a large number of complex JEXL expressions.

bigcyy: >Then I looked at the code you discussed for this position. As I understand it, you should first perform a QueryExpr query, then use expr to perform JEXL threshold matching. Wouldn't this be a bit complicated? I'm also concerned that there might be a performance bottleneck when executing a large number of complex JEXL expressions.

I believe this is unavoidable. We can, however, add expression caching or SQL result set caching to improve performance.

bigcyy: Furthermore, when coupling SQL with a threshold expression, the SQL result must be a single column; a multi-column result set cannot be processed. For example, if I need to count the number of logs for each log level of every web service within 30 seconds, and this SQL is coupled with a threshold expression, the result will only contain the log count column. This leads to a loss of critical information, such as which services and which log levels the logs belong to.

Duansg:
Hi, @bigcyy I pulled your PR and gave it a quick spin. Overall, aside from some UI interactions that need polishing ,it's pretty well-rounded. I'd be happy to be the first user after this PR merges, hahaha. :+1:

Maybe we can even build this feature together :)

Here are some of my questions and suggestions:

1. Without Greptime enabled, can the `Log Manage` list page display a user-friendly prompt instead of an error message?
2. In the Threshold configuration, the `Alarm Content` now displays in a single column. Was this intentional?
3. In the `Periodic Threshold` configuration, do both metrics and logs now support PROMQL and SQL? Because I noticed that when using SQL queries for metrics, they failed to display properly.
4. Could we consider initializing the `Log Stream` in a `pause` state? When `resume` is triggered, it would then initialize the connection state. Additionally, should we adjust the automatic refresh of input parameters to: take effect only when manually paused/enabled? Currently, when I input filtering parameters, if I haven't finished entering them before resetting the SSE, I lose the reference logs.

hertzbeat-common/src/main/java/org/apache/hertzbeat/common/entity/alerter/AlertDefine.java
tomsun28: I wonder if we are using JEXL here, which leads to the separation of the two expressions now. What if the SQL query itself is an expression judgment? I see that some platforms do this.  

> For example, if I need to count the number of logs for each log level of every web service within 30 seconds which number is > 20.

For this, the sql expr is 
```
SELECT 
    service_name,
    log_level,
    COUNT(*) AS log_count
FROM logs
WHERE timestamp >= NOW() - INTERVAL '30' SECOND
GROUP BY service_name, log_level
HAVING COUNT(*) > 20;
```
That we do not need other expr to condition. Also it can render the filter data UI.

tomsun28: Why recommend using a single expression instead of two

- Ease of sharing and dissemination, similar to a PromQL threshold expression or an expression like `{level="error"} |~ ".*(timeout|exception).*"` or a sql, which makes it very convenient to share and propagate.
- Unified design, keeping it consistent with the design of other existing types of expressions, where the expression itself includes both the data query and the trigger condition. like the PromQL、Loki、ELK.
- For future unification of expressions, we can consider defining our own unified expression rules for external use, and internally implement them based on different dependencies.
- AI can better understand and generate expressions throught one expr.

bigcyy: Hi Tom,

I understand the alerting capabilities of Alibaba Cloud's SLS, particularly why two expressions are needed for advanced scenarios.
- Setting multi-level alerts: Binding different threshold expressions to various severity levels (e.g., critical, warning) based on the same query result.
- Conducting visual analysis: Creating complex visualizations from the query results before alerting.
- Creating composite alerts: Combining results from multiple, separate queries into a single, consolidated alert.

However, after carefully reviewing our specific requirements, I've confirmed that our current alerting logic doesn't involve these complexities. Our needs can be fully and efficiently met with just a single expression.

To keep our configuration streamlined and better aligned with our use case, I will proceed with modifying our setup to use a single-expression alert.

bigcyy:
> Hi, @bigcyy I pulled your PR and gave it a quick spin. Overall, aside from some UI interactions that need polishing ,it's pretty well-rounded. I'd be happy to be the first user after this PR merges, hahaha. 👍
> 
> Maybe we can even build this feature together :)
> 
> Here are some of my questions and suggestions:
> 
> 1. Without Greptime enabled, can the `Log Manage` list page display a user-friendly prompt instead of an error message?
> 2. In the Threshold configuration, the `Alarm Content` now displays in a single column. Was this intentional?
> 3. In the `Periodic Threshold` configuration, do both metrics and logs now support PROMQL and SQL? Because I noticed that when using SQL queries for metrics, they failed to display properly.
> 4. Could we consider initializing the `Log Stream` in a `pause` state? When `resume` is triggered, it would then initialize the connection state. Additionally, should we adjust the automatic refresh of input parameters to: take effect only when manually paused/enabled? Currently, when I input filtering parameters, if I haven't finished entering them before resetting the SSE, I lose the reference logs.

Hi @Duansg , Thanks so much for the detailed feedback and for taking the time to test out my PR. I'm glad to hear you found it well-rounded. It would be an honor to have you as the first user!

I appreciate the offer to collaborate, but as this is part of my  OSPP activity, I need to complete this PR by myself. After it's merged into the main branch, I would be very happy to collaborate on future improvements based on your interest.

Here are my thoughts on your questions and suggestions:

- Log Management Page Prompt: This is an excellent suggestion. I agree that a user-friendly prompt would be much better than an error message, and I'll make sure to address this in a future update. https://github.com/apache/hertzbeat/pull/3673/commits/729426f93ca00681384d69e23fe90bcb8f43c5e6
- Alarm Content Display: The **Alarm Content** is designed to display as a grouped alert by default, which is why it shows in a single column.
- Periodic Threshold Queries: Currently, logs don't support PromQL. I'll make sure the front end is updated to reflect this limitation. Metrics should support SQL queries, so I'll investigate why they failed to display properly. Thanks for pointing that out! https://github.com/apache/hertzbeat/pull/3673/commits/5489a71b03add93e6c0d46f9523a53927f6613b3
- Log Stream State: You're suggesting that the Log Stream should be paused by default when you first enter the page. Is that correct? I think that's a great idea. As for the filtering parameters, what if we changed it so that the filtering only takes effect when you click a button? That way, you'd have time to finish typing before the SSE connection resets. https://github.com/apache/hertzbeat/pull/3673/commits/5489a71b03add93e6c0d46f9523a53927f6613b3

Thanks again for all your help and for taking the time to review my PR.

hertzbeat-common/src/main/java/org/apache/hertzbeat/common/entity/alerter/AlertDefine.java
bigcyy: https://github.com/apache/hertzbeat/pull/3673/commits/bc4cfa1933608bea1c244be5e10e76297245c07f

Duansg:
Hi, @bigcyy by the way, here are some suggestions you might find helpful:

1. Optimization: Currently, entering an excessively large value in the page settings window triggers an error due to missing validation here.
2. Recommendation: Limit the window size (`windowSizeMs`). Currently, it appears to accept the maximum value for an int. Setting the window size excessively large not only increases memory pressure but may also result in a large number of "lively" windows persisting in the system. This could subsequently impact the processing efficiency of `onWatermark`.
3. Recommendation: Limit the size of `activeWindows`. In certain scenarios (e.g., a large number of distinct `alertDefine` instances or extremely large window size settings), `activeWindows` may grow excessively large, potentially leading to memory leaks in extreme cases.
4. Recommendations: Implement fault tolerance or [monitoring](https://github.com/apache/hertzbeat/pull/3641) for certain critical resource usage. For example, in the processing logic of `org.apache.hertzbeat.alert.calculate.realtime.window.AlarmEvaluator#workerExecutor`, if rejected, it may result in the loss of entire window data, and I cannot dynamically adjust the thread pool size.

Duansg:
> Hi, @bigcyy by the way, here are some suggestions you might find helpful:
> 
> 1. Optimization: Currently, entering an excessively large value in the page settings window triggers an error due to missing validation here.
> 2. Recommendation: Limit the window size (`windowSizeMs`). Currently, it appears to accept the maximum value for an int. Setting the window size excessively large not only increases memory pressure but may also result in a large number of "lively" windows persisting in the system. This could subsequently impact the processing efficiency of `onWatermark`.
> 3. Recommendation: Limit the size of `activeWindows`. In certain scenarios (e.g., a large number of distinct `alertDefine` instances or extremely large window size settings), `activeWindows` may grow excessively large, potentially leading to memory leaks in extreme cases.
> 4. Recommendations: Implement fault tolerance or [monitoring](https://github.com/apache/hertzbeat/pull/3641) for certain critical resource usage. For example, in the processing logic of `org.apache.hertzbeat.alert.calculate.realtime.window.AlarmEvaluator#workerExecutor`, if rejected, it may result in the loss of entire window data, and I cannot dynamically adjust the thread pool size.

5.The entire mechanism currently relies on the normal generation and advancement of `Watermark`. The generation method for `newWatermark` is `maxTs - DEFAULT_WATERMARK_DELAY_MS`, while `maxTs` indirectly depends on the time of newly input `LogEntry`. If events sent by systems with unsynchronized clocks have timestamps ahead of schedule, could this cause the watermark to stagnate and result in delayed alerts?

bigcyy:
> 5.The entire mechanism currently relies on the normal generation and advancement of `Watermark`. The generation method for `newWatermark` is `maxTs - DEFAULT_WATERMARK_DELAY_MS`, while `maxTs` indirectly depends on the time of newly input `LogEntry`. If events sent by systems with unsynchronized clocks have timestamps ahead of schedule, could this cause the watermark to stagnate and result in delayed alerts?

You are absolutely right, thank you for pointing this out. This is a critical issue that I completely overlooked in my code.

You've correctly identified a serious flaw in my logic. When an event with a timestamp far in the future is processed, my current implementation would skew the `maxTs`. This, in turn, causes the `newWatermark` (calculated as `maxTs - DEFAULT_WATERMARK_DELAY_MS`) to jump far ahead of the actual event time. The result is that the watermark would stagnate because normally timed events would be incorrectly considered "late" and fail to advance it, leading to significant delays in the alerting system.

To fix this, I plan to implement a "stagnation detection and forced advancement" mechanism. Here is the solution I will be coding:

1.  Track the Last Watermark Update Time: I will add a new variable, `lastWatermarkUpdateTime`, to record the timestamp of the last successful watermark update.

2.  Check for Stagnation: Within the `TimeService`'s regular watermark check, if the newly calculated watermark is the same as the existing one, the logic will treat this as a potential stagnation.

3.  Measure Stagnation Duration: If no update occurs, I'll add a step to calculate the stagnation period by subtracting `lastWatermarkUpdateTime` from the current system time.

4.  Force a Watermark Update: If this stagnation period exceeds a predefined threshold (e.g., `MAX_WATERMARK_STAGNATION_TIME`), my code will determine that the watermark has been stuck for too long. At this point, instead of relying on the potentially incorrect `maxTs`, it will forcibly advance the watermark to `now() - DEFAULT_WATERMARK_DELAY_MS`. This will ensure the processing pipeline moves forward, even when compromised by future-dated events.

Thank you again for raising this. Your insight is invaluable for helping me improve the robustness and reliability of my code. I will prioritize implementing this fix.

Duansg:
> > 5.The entire mechanism currently relies on the normal generation and advancement of `Watermark`. The generation method for `newWatermark` is `maxTs - DEFAULT_WATERMARK_DELAY_MS`, while `maxTs` indirectly depends on the time of newly input `LogEntry`. If events sent by systems with unsynchronized clocks have timestamps ahead of schedule, could this cause the watermark to stagnate and result in delayed alerts?
> 
> You are absolutely right, thank you for pointing this out. This is a critical issue that I completely overlooked in my code.
> 
> You've correctly identified a serious flaw in my logic. When an event with a timestamp far in the future is processed, my current implementation would skew the `maxTs`. This, in turn, causes the `newWatermark` (calculated as `maxTs - DEFAULT_WATERMARK_DELAY_MS`) to jump far ahead of the actual event time. The result is that the watermark would stagnate because normally timed events would be incorrectly considered "late" and fail to advance it, leading to significant delays in the alerting system.
> 
> To fix this, I plan to implement a "stagnation detection and forced advancement" mechanism. Here is the solution I will be coding:
> 
> 1. Track the Last Watermark Update Time: I will add a new variable, `lastWatermarkUpdateTime`, to record the timestamp of the last successful watermark update.
> 2. Check for Stagnation: Within the `TimeService`'s regular watermark check, if the newly calculated watermark is the same as the existing one, the logic will treat this as a potential stagnation.
> 3. Measure Stagnation Duration: If no update occurs, I'll add a step to calculate the stagnation period by subtracting `lastWatermarkUpdateTime` from the current system time.
> 4. Force a Watermark Update: If this stagnation period exceeds a predefined threshold (e.g., `MAX_WATERMARK_STAGNATION_TIME`), my code will determine that the watermark has been stuck for too long. At this point, instead of relying on the potentially incorrect `maxTs`, it will forcibly advance the watermark to `now() - DEFAULT_WATERMARK_DELAY_MS`. This will ensure the processing pipeline moves forward, even when compromised by future-dated events.
> 
> Thank you again for raising this. Your insight is invaluable for helping me improve the robustness and reliability of my code. I will prioritize implementing this fix.

@bigcyy hhh, you're welcome. I'm glad I could help you.

bigcyy:
> > 5.The entire mechanism currently relies on the normal generation and advancement of `Watermark`. The generation method for `newWatermark` is `maxTs - DEFAULT_WATERMARK_DELAY_MS`, while `maxTs` indirectly depends on the time of newly input `LogEntry`. If events sent by systems with unsynchronized clocks have timestamps ahead of schedule, could this cause the watermark to stagnate and result in delayed alerts?
> 
> You are absolutely right, thank you for pointing this out. This is a critical issue that I completely overlooked in my code.
> 
> You've correctly identified a serious flaw in my logic. When an event with a timestamp far in the future is processed, my current implementation would skew the `maxTs`. This, in turn, causes the `newWatermark` (calculated as `maxTs - DEFAULT_WATERMARK_DELAY_MS`) to jump far ahead of the actual event time. The result is that the watermark would stagnate because normally timed events would be incorrectly considered "late" and fail to advance it, leading to significant delays in the alerting system.
> 
> To fix this, I plan to implement a "stagnation detection and forced advancement" mechanism. Here is the solution I will be coding:
> 
> 1. Track the Last Watermark Update Time: I will add a new variable, `lastWatermarkUpdateTime`, to record the timestamp of the last successful watermark update.
> 2. Check for Stagnation: Within the `TimeService`'s regular watermark check, if the newly calculated watermark is the same as the existing one, the logic will treat this as a potential stagnation.
> 3. Measure Stagnation Duration: If no update occurs, I'll add a step to calculate the stagnation period by subtracting `lastWatermarkUpdateTime` from the current system time.
> 4. Force a Watermark Update: If this stagnation period exceeds a predefined threshold (e.g., `MAX_WATERMARK_STAGNATION_TIME`), my code will determine that the watermark has been stuck for too long. At this point, instead of relying on the potentially incorrect `maxTs`, it will forcibly advance the watermark to `now() - DEFAULT_WATERMARK_DELAY_MS`. This will ensure the processing pipeline moves forward, even when compromised by future-dated events.
> 
> Thank you again for raising this. Your insight is invaluable for helping me improve the robustness and reliability of my code. I will prioritize implementing this fix.

I've decided not to use the previously mentioned solution because it doesn't fix the root problem.

Instead, I will use the local system clock to maintain an acceptable time range for logs. This approach ensures that logs with anomalous timestamps cannot incorrectly advance the watermark.

Under this new logic, logs with timestamps below the watermark will be dropped, while logs above the watermark will only be processed if their timestamps are within this reasonable range.

zhangshenghang:
Please provide a separate document to explain how to use this feature, preferably accompanied by relevant images to guide users.

bigcyy:
> Please provide a separate document to explain how to use this feature, preferably accompanied by relevant images to guide users.

done

Issue #3772, opened 2025-09-17
[BUG] <errorlog> From time to time, errors can be observed in the log files.


love01211:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

There are no issues with the interface or functionality, but when I reviewed the backend logs, I found error logs with no discernible pattern in their timing.

### Steps To Reproduce

1. cat localhost.localdomain2540-2025-09-17-error.0.log
2025-09-17 13:58:00.796 [http-nio-1157-exec-1] ERROR org.apache.hertzbeat.manager.support.GlobalExceptionHandler Line:252 - [monitor]-[unknown error happen]-Response not usable after response errors.
org.springframework.web.context.request.async.AsyncRequestNotUsableException: Response not usable after response errors.
        at org.springframework.web.context.request.async.StandardServletAsyncWebRequest$LifecycleHttpServletResponse.obtainLockOrRaiseException(StandardServletAsyncWebRequest.java:338)
        at org.springframework.web.context.request.async.StandardServletAsyncWebRequest$LifecycleHttpServletResponse.flushBuffer(StandardServletAsyncWebRequest.java:320)
        at org.springframework.http.server.ServletServerHttpResponse.flush(ServletServerHttpResponse.java:105)
        at org.springframework.http.server.DelegatingServerHttpResponse.flush(DelegatingServerHttpResponse.java:61)
        at org.springframework.web.servlet.mvc.method.annotation.ResponseBodyEmitterReturnValueHandler$DefaultSseEmitterHandler.complete(ResponseBodyEmitterReturnValueHandler.java:320)
        at org.springframework.web.servlet.mvc.method.annotation.ResponseBodyEmitter.complete(ResponseBodyEmitter.java:252)
        at org.apache.hertzbeat.alert.config.AlertSseManager.lambda$broadcast$3(AlertSseManager.java:57)
        at java.base/java.util.concurrent.ConcurrentHashMap.forEach(ConcurrentHashMap.java:1603)
        at org.apache.hertzbeat.alert.config.AlertSseManager.broadcast(AlertSseManager.java:50)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
        at java.base/java.lang.reflect.Method.invoke(Method.java:580)
        at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:359)
[...]

PR #3775, opened 2025-09-17
[improve] SSE exception handling improvements

Duansg:
## What's changed?

improve #3772 

Ignore exceptions occurring when the SSE connection is terminated, this HTTP response has entered an error state or been closed and can no longer be manipulated with methods such as send or complete.

1. If an exception occurs during `emitter.complete()`, it may prevent the existing connections from being cleared, potentially leading to a memory leak.
2. In this case, the exception merely wraps the IOException present in the response. In other words, the exception already exists, and it indicates that the response is no longer available and cannot send any further data.

## Checklist

## Add or update API


MasamiYui:
Hi, ManagerSseManager also needs to be modified

Duansg:
> Hi, ManagerSseManager also needs to be modified

Got it. :thumbsup:

MasamiYui (approved):
LGTM.

PR #3797, opened 2025-09-29
[feat] Add webhook authentication support for alert notifications

Jetiaime:
- Add Basic Auth and Bearer Token authentication types to webhook notifications
- Update NoticeReceiver entity with hookAuthType and hookAuthToken fields
- Implement authentication headers in WebHookAlertNotifyHandlerImpl
- Add frontend UI support for webhook authentication configuration
- Update i18n translations for new authentication options

## What's changed?

Add Authentication Header On WebHooks

## Checklist

## Add or update API


Jetiaime:
> Hi, I have a suggestion to discuss with you, This implementation maybe too specific. how about allowing users to set arbitrary http headers, which can cover more authorizations like Basic/Bearer or company-specific authorization headers.

Yeah, I consider about it, but there are some problems we need to solve first: If we allow users to pass the whole headers, it may passed it as a HashMap, there is not have a good way to render it in the frontend. I prefer a simple feature for now, to keep the authentication part simple (user may enter a whole json string or add key-value table like the notifier rules, there is more complexity). And because I'm new to this project, So I wanna to push a simple feature first, then we can discuss more about the headers part. What do you think? If you can help to implement it, that would be great.

MasamiYui:
> > Hi, I have a suggestion to discuss with you, This implementation maybe too specific. how about allowing users to set arbitrary http headers, which can cover more authorizations like Basic/Bearer or company-specific authorization headers.
> 
> Yeah, I consider about it, but there are some problems we need to solve first: If we allow users to pass the whole headers, it may passed it as a HashMap, there is not have a good way to render it in the frontend. I prefer a simple feature for now, to keep the authentication part simple (user may enter a whole json string or add key-value table like the notifier rules, there is more complexity). And because I'm new to this project, So I wanna to push a simple feature first, then we can discuss more about the headers part. What do you think? If you can help to implement it, that would be great.

Hi, about UI design of HTTP headers, you can either refer to the designs of other systems or implement it according to your own ideas. If any issues, we could discuss together.

tomsun28:
Hi👍, suggest the common HTTP headers +1.  
About the ```map render in the front ui```, how about use the https://github.com/apache/hertzbeat/tree/master/web-app/src/app/shared/components/configurable-field

Aias00:
@Jetiaime hi, pls fix ci

eunrui:
really need this feature

Jetiaime:
I'm sorry I haven't been able to handle this PR for a while due to my work commitments. I have considered another approach to add more common headers in the new PR. This will be a temporary solution. Please approve this PR first, and I will add the new common headers in the next few weeks.

tomsun28:
👍 It’s OK. Take your time.
Please fix the webapp eslint error, due the ci not paas, we cannot merge it.

PR #3803, opened 2025-10-05
fix: add use soft deletion + scheduled tasks for monitoring

### LINKED ISSUE #3798: A very significant bug: when you keep refreshing, the monitoring entries will persist and then disappear automatically.

### Current Behavior

Regarding the monitoring entries of DNS service discovery, when you keep refreshing, the entries will appear from time to time. However, when you click to enter and edit them, a prompt saying "the entry does not exist" sometimes appears. The entries will disappear automatically and then reappear unexpectedly.
Please refer to the screen recording for specific information：
https://github.com/user-attachments/assets/9650a109-a88b-4357-a2ec-734ee50c652e

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
HertzBeat version(s): 1.7.3
Docker
```

### Debug logs

_No response_

### Anything else?

_No response_

mengnankkkk:
## What's changed?
Close #3798

Direct physical deletion of monitoring items that are not in the service discovery data causes a problem of being out of sync with the front-end acquisition status. The front-end does not handle the error situation where the monitoring items have been deleted. I will use soft deletion + scheduled tasks to solve this problem

## Checklist

## Add or update API

### DIFF
--- a/hertzbeat-collector/hertzbeat-collector-basic/src/main/java/org/apache/hertzbeat/collector/collect/prometheus/parser/OnlineParser.java
+++ b/hertzbeat-collector/hertzbeat-collector-basic/src/main/java/org/apache/hertzbeat/collector/collect/prometheus/parser/OnlineParser.java
@@ -352,12 +352,47 @@ public class OnlineParser {
                         }
                     }
                 }
-            } else {
+            } else if (i <= 127) {
                 stringBuilder.append((char) i);
+            } else {
+                handleUtf8Character(i, inputStream, stringBuilder);
             }
             i = inputStream.read();
         }
         return new CharChecker(i);
+    }
+
+    private static void handleUtf8Character(int firstByte, InputStream inputStream, StringBuilder stringBuilder) throws IOException {
+        List<Integer> bytes = new ArrayList<>();
+        bytes.add(firstByte);
+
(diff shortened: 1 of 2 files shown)


pentium100:
I think this is not a bug. The delete action of sub monitor is success. and then the service discovery worker is running again and create a new sub monitor. if user really want to delete this sub monitor, user need to delete service discovery monitor or service discovery will re-create it again.

mengnankkkk:
> I think this is not a bug. The delete action of sub monitor is success. and then the service discovery worker is running again and create a new sub monitor. if user really want to delete this sub monitor, user need to delete service discovery monitor or service discovery will re-create it again.

I want to optimize this problem from the perspective of user experience. Flickering on the monitor may cause users to have doubts. You can adopt my strategy. If you think it is unnecessary to make such a modification, I can only modify the front end to avoid this problem.

tomsun28:
👍Hi thanks for contribution, it is not suggest to use the soft deletion for tasks. How about just modify the front end to fix it as you metion? thanks 👍

mengnankkkk:
> 👍Hi thanks for contribution, it is not suggest to use the soft deletion for tasks. How about just modify the front end to fix it as you metion? thanks 👍
> 
> 
Ok  I will do it tomorrow 💕💕

PR #3810, opened 2025-10-09
[refactor]Refactoring prometheus label value utf8 support

Duansg:
## What's changed?

refactor #3792

Refactor and optimized support for UTF-8 encoding in certain Prometheus label values.

1. Optimize fixed-size arrays to avoid dynamic allocation and prevent unnecessary memory consumption.
2. Add the replacement character `\uFFFD` when handling UTF-8 errors to comply with standards.
3. Added basic over-encoding checks and format validation.
4. Optimize exception handling and subsequent error validation.
5. Optimized certain processing logic to reduce unnecessary operations for early validation.
6. Passed historical and newly added test cases

Reference:
1. https://datatracker.ietf.org/doc/html/rfc3629
2. https://en.wikipedia.org/wiki/UTF-8

## Checklist

## Add or update API

> Monitors
Snipaste_2025-10-09_15-46-59

> Alarms
Snipaste_2025-10-09_15-58-43

### DIFF
--- a/hertzbeat-collector/hertzbeat-collector-basic/src/test/java/org/apache/hertzbeat/collector/collect/prometheus/parser/OnlineParserSingleTest.java
+++ b/hertzbeat-collector/hertzbeat-collector-basic/src/test/java/org/apache/hertzbeat/collector/collect/prometheus/parser/OnlineParserSingleTest.java
@@ -489,7 +489,7 @@ public class OnlineParserSingleTest {
 
     @Test
     void testHandleUtf8Character_ValidThreeByteSequence() throws Exception {
-        // Test valid 3-byte UTF-8 sequence: 中 (U+4E2D) = 0xE4 0xB8 0xAD
+        // Test valid 3-byte UTF-8 sequence: Chinese character (U+4E2D) = 0xE4 0xB8 0xAD
         String str = "test_metric{label=\"\u4e2d\u6587\"} 1\n";
         InputStream inputStream = new ByteArrayInputStream(str.getBytes(StandardCharsets.UTF_8));
         Map<String, MetricFamily> metricFamilyMap = parseMetrics(inputStream, "test_metric");
@@ -501,7 +501,7 @@ public class OnlineParserSingleTest {
         
         MetricFamily.Label label = metricFamily.getMetricList().get(0).getLabels().get(0);
         assertEquals("label", label.getName());
-        assertEquals("中文", label.getValue());
+        assertEquals("\u4e2d\u6587", label.getValue());
     }
 
     @Test
@@ -596,6 +596,6 @@ public class OnlineParserSingleTest {
         assertEquals("label", label.getName());
         String value = label.getValue();
         assertNotNull(value);
-        assertEquals("Helloé中😀", value);
+        assertEquals("Hello\u00e9\u4e2d\ud83d\ude00", value);
     }
 }


tomsun28:
👍 relate https://github.com/apache/hertzbeat/pull/3792

tomsun28:
ptal @mengnankkkk thanks

hertzbeat-collector/hertzbeat-collector-basic/src/main/java/org/apache/hertzbeat/collector/collect/prometheus/parser/OnlineParser.java
```diff
@@ -362,37 +365,123 @@ private static CharChecker parseLabelValue(InputStream inputStream, StringBuilde
...
+        for (int i = 0; i < additionalBytes; i++) {
+            int nextByte = inputStream.read();
+            if (nextByte == -1) {
+                appendInvalidCharacters(stringBuilder);
+                return;
+            }
+            // Verify subsequent byte format:10xxxxxx
+            if ((nextByte & 0xC0) != 0x80) {
+                appendInvalidCharacters(stringBuilder);
+                return;
```
mengnankkkk, line 397: There seems to be a lack of surrogate and U+10FFFF checks here ❤

Duansg: > There seems to be a lack of surrogate and U+10FFFF checks here ❤

@mengnankkkk  Hi, thank you for your reply. Regarding the point you raised, I have indeed considered it.

My consideration is:
1. It cannot appear in a standard UTF-8 data stream, as RFC 3629 explicitly prohibits it.
2. I reviewed the documentation related to Prometheus. It only specifies UTF-8 encoding but does not describe any associated code point detection.
3. The `label value` generated by the official Prometheus exporter will never appear, as it is considered invalid UTF-8 within the official Prometheus ecosystem (including the Go implementation and mainstream exporters) or when using the Go standard library. Consequently, it will never appear in valid metric outputs.

Therefore, a relatively lenient approach was adopted. Such scenarios are extremely rare and only occur in non-standard implementations or when error byte streams are intentionally generated manually. Consequently, I added a `todo` annotation in the code to ensure additional detection and handling can be implemented if necessary in the future.

What do you think?

mengnankkkk: Ok, your idea is correct. I agree with your point of view and the next step.💕💕

Duansg:
@mengnankkkk Sorry I forgot to reply to your question in #3811. When handling this part of the logic, my thought was:

Because the start byte does not meet the rules, returning directly would leave residual continuation bytes. However, the unknown replacement character has been added because I modified the `isInvalidLabelValue` method to check for unknown replacement characters. These are validated and interrupted (throwing an exception). While trailing continuation bytes would continue parsing, this is limited to the currently parsed label value. Therefore, no additional complex handling for trailing bytes is added unless future implementation requires correctly parsing the current indicator data—unlike the current unacceptable state.

Although I've considered these points, I'm also very interested in hearing your thoughts.

mengnankkkk:
> @mengnankkkk Sorry I forgot to reply to your question in #3811. When handling this part of the logic, my thought was:
> 
> Because the start byte does not meet the rules, returning directly would leave residual continuation bytes. However, the unknown replacement character has been added because I modified the `isInvalidLabelValue` method to check for unknown replacement characters. These are validated and interrupted (throwing an exception). While trailing continuation bytes would continue parsing, this is limited to the currently parsed label value. Therefore, no additional complex handling for trailing bytes is added unless future implementation requires correctly parsing the current indicator data—unlike the current unacceptable state.
> 
> Although I've considered these points, I'm also very interested in hearing your thoughts.

OK, we can add a comment to the isInvalidLabelValue method to explain why we do this. I see you have already done it. I think this solution is still very good and it solves the problem simply! Thank you for your reply

PR #3848, opened 2025-11-07
[bugfix]manager: Avoid NPE when metrics data queue is empty

starryCoder:
## What's changed?

Fixes a potential NullPointerException when `dataQueue.pollServiceDiscoveryData()` and `dataQueue.pollMetricsDataToAlerter()` returns null due to an empty queue or non-blocking behavior.

The `metricsData` object was not checked for null before calling `metricsData.getId()`, leading to continuous NPE errors in the log.

## Checklist

## Add or update API

### DIFF
--- a/hertzbeat-common/src/main/java/org/apache/hertzbeat/common/queue/impl/RedisCommonDataQueue.java
+++ b/hertzbeat-common/src/main/java/org/apache/hertzbeat/common/queue/impl/RedisCommonDataQueue.java
@@ -17,6 +17,7 @@
 
 package org.apache.hertzbeat.common.queue.impl;
 
+import io.lettuce.core.KeyValue;
 import io.lettuce.core.RedisClient;
 import io.lettuce.core.RedisURI;
 import io.lettuce.core.api.StatefulRedisConnection;
@@ -86,33 +87,19 @@ public class RedisCommonDataQueue implements CommonDataQueue, DisposableBean {
     }
 
     @Override
-    public CollectRep.MetricsData pollMetricsDataToAlerter() {
-        try {
-            return syncCommands.rpop(metricsDataQueueNameToAlerter);
-        } catch (Exception e) {
-            log.error(e.getMessage());
-            return null;
-        }
+    public CollectRep.MetricsData pollMetricsDataToAlerter() throws InterruptedException {
+        return genericBlockingPollFunction(metricsDataQueueNameToAlerter, syncCommands);
     }
 
     @Override
     public CollectRep.MetricsData pollMetricsDataToStorage() throws InterruptedException {
-        try {
-            return syncCommands.rpop(metricsDataQueueNameToStorage);
-        } catch (Exception e) {
-            log.error(e.getMessage());
-            return null;
-        }
+        return genericBlockingPollFunction(metricsDataQueueNameToStorage, syncCommands);
+
     }
 
     @Override
     public CollectRep.MetricsData pollServiceDiscoveryData() throws InterruptedException {
-        try {
-            return syncCommands.rpop(metricsDataQueueNameForServiceDiscovery);
-        } catch (Exception e) {
-            log.error(e.getMessage());
-            return null;
-        }
+        return genericBlockingPollFunction(metricsDataQueueNameForServiceDiscovery, syncCommands);
     }
 
     @Override
@@ -153,12 +140,7 @@ public class RedisCommonDataQueue implements CommonDataQueue, DisposableBean {
 
     @Override
     public LogEntry pollLogEntry() throws InterruptedException {
-        try {
-            return logEntrySyncCommands.rpop(logEntryQueueName);
-        } catch (Exception e) {
-            log.error("Failed to poll LogEntry from Redis: {}", e.getMessage());
-            throw new InterruptedException("Failed to poll LogEntry from Redis");
-        }
+        return genericBlockingPollFunction(logEntryQueueName, logEntrySyncCommands);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


hertzbeat-alerter/src/main/java/org/apache/hertzbeat/alert/calculate/realtime/MetricsRealTimeAlertCalculator.java
```diff
@@ -125,6 +125,9 @@ public void startCalculate() {
             while (!Thread.currentThread().isInterrupted()) {
                 try {
                     CollectRep.MetricsData metricsData = dataQueue.pollMetricsDataToAlerter();
+                    if (metricsData == null) {
```
Duansg, line 128: Good question. However, in a non-blocking implementation, this would create a classic **busy-waiting** loop.

Duansg:
Consider replacing the **blocking wait** approach with an alternative, such as: Redis rpop → brpop

starryCoder:
> Consider replacing the **blocking wait** approach with an alternative, such as: Redis rpop → brpop

ok, let me try.

tomsun28:
> > Consider replacing the **blocking wait** approach with an alternative, such as: Redis rpop → brpop
> 
> ok, let me try.

i have a confuse, we currently have use the BlockingQueue.take() - the blocking implementation -> like brpop

Duansg:
> i have a confuse, we currently have use the BlockingQueue.take() - the blocking implementation -> like brpop

He must have used some other non-blocking queue implementation. :joy:

starryCoder:
> > > Consider replacing the **blocking wait** approach with an alternative, such as: Redis rpop → brpop
> > 
> > 
> > ok, let me try.
> 
> i have a confuse, we currently have use the BlockingQueue.take() - the blocking implementation -> like brpop

I pulled the latest master code to address the memory leak issue, but now I'm running into a NullPointerException (NPE) when trying to use both Redis and Kafka.

Duansg:
Regarding the retry strategy here, an exponential backoff strategy can be implemented on the consumer side to prevent rapid failures caused by unknown exceptions in the queue implementation strategy from resulting in busy waiting on the consumer side.

I will attempt to modify and enhance this strategy in subsequent iterations.

@tomsun28 @starryCoder Do you have any suggestions regarding the improvements I mentioned above?

starryCoder:
> Regarding the retry strategy here, an exponential backoff strategy can be implemented on the consumer side to prevent rapid failures caused by unknown exceptions in the queue implementation strategy from resulting in busy waiting on the consumer side.
> 
> I will attempt to modify and enhance this strategy in subsequent iterations.
> 
> @tomsun28 @starryCoder Do you have any suggestions regarding the improvements I mentioned above?

If the BRPOP timeout is set to 0 (infinite block), the consumer is only unblocked when a message arrives or when a connection/network failure occurs. Therefore, the Exponential Backoff logic for normal queue emptiness is indeed unnecessary.

Duansg:
> > Regarding the retry strategy here, an exponential backoff strategy can be implemented on the consumer side to prevent rapid failures caused by unknown exceptions in the queue implementation strategy from resulting in busy waiting on the consumer side.
> > I will attempt to modify and enhance this strategy in subsequent iterations.
> > @tomsun28 @starryCoder Do you have any suggestions regarding the improvements I mentioned above?
> 
> If the BRPOP timeout is set to 0 (infinite block), the consumer is only unblocked when a message arrives or when a connection/network failure occurs. Therefore, the Exponential Backoff logic for normal queue emptiness is indeed unnecessary.

That's not what I meant. What I wanted to describe was:

For example: in `RedisCommonDataQueue`

```
try {
   return syncCommands.rpop(metricsDataQueueNameToAlerter);
} catch (Exception e) {
   return null;
}
```
If a Redis call causes the connection to break due to an unknown exception, an exception will be thrown immediately and null will be returned. It checks for an empty condition externally and proceeds to the next iteration of the loop.

Duansg:
This is also one of the reasons why an exception is thrown when retrieving log data, thereby interrupting the outer loop.

```java
public LogEntry pollLogEntry() throws InterruptedException {
  try {
    return logEntrySyncCommands.rpop(logEntryQueueName);
  } catch (Exception e) {
    log.error("Failed to poll LogEntry from Redis: {}", e.getMessage());
    throw new InterruptedException("Failed to poll LogEntry from Redis");
  }
}
```

@bigcyy Is my understanding correct? Could you take a look?

bigcyy:
> @bigcyy Is my understanding correct? Could you take a look?

Thank you for clarifying. I just want to summarize my understanding of the proposed improvements to make sure I'm on the same page.
As I understand it, the two main suggestions are:
1. Replace RPOP with BRPOP: The goal here is to avoid the consumer thread from busy-waiting or spinning in a tight loop when the queue is empty. Using the blocking BRPOP (with an appropriate timeout) is much more efficient and prevents unnecessary CPU waste.
2. Implement Exponential Backoff for Exceptions: When the consumer encounters an unexpected exception (such as a network error or a connection failure, rather than just an empty queue), the thread should not terminate. Instead, it should implement an exponential backoff strategy for retrying the connection/operation. This makes the consumer more resilient and prevents it from either crashing or overwhelming the system with rapid retries during a failure.


Is my understanding of your proposed solutions correct?

Duansg:
> Is my understanding of your proposed solutions correct?

:+1: Yes, your summary is correct. My idea is that adding an exponential backoff strategy at the public consumption end, without concerning the implementation of the queue, won't affect the overall logic. It only needs to be triggered when the return value is null, for example: 50ms, 100ms, 500ms...1s.  This prevents the queue implementation from causing system overload due to exceptionally rapid failures.

starryCoder:
> > Is my understanding of your proposed solutions correct?
> 
> 👍 Yes, your summary is correct. My idea is that adding an exponential backoff strategy at the public consumption end, without concerning the implementation of the queue, won't affect the overall logic. It only needs to be triggered when the return value is null, for example: 50ms, 100ms, 500ms...1s. This prevents the queue implementation from causing system overload due to exceptionally rapid failures.

That makes perfect sense! This strategy is much cleaner and more robust.

PR #3854, opened 2025-11-15
fix: Remove redundant metrics data send in KafkaCommonDataQueue

starryCoder:
The implementation of sendMetricsDataToStorage was redundant. Both the alert consumer and the storage consumer subscribe to the same metricsDataTopic using different group IDs.

## What's changed?
In the current Kafka queue setup, the following consumers are configured to handle metrics data:
1. The Alert Consumer `metricsDataToAlertConsumer` subscribes to the `metricsDataTopic`.
2. The Storage Consumer `metricsDataToStorageConsumer` is configured with a different group.id `metrics-persistent-consumer` and also subscribes to the same `metricsDataTopic`.

Due to Kafka's publish-subscribe model, any message sent to `metricsDataTopic` is already consumed by both the Alert and Storage consumer groups.
The original implementation of `sendMetricsDataToStorage` was redundant, as it attempted to send the same `MetricsData` to a different storage-specific topic `metricsDataToStorageTopic`

so, The `sendMetricsDataToStorage` method is changed to an empty implementation .
## Checklist

## Add or update API


hertzbeat-common/src/main/java/org/apache/hertzbeat/common/queue/impl/KafkaCommonDataQueue.java
```diff
@@ -195,13 +195,8 @@ public void sendMetricsData(CollectRep.MetricsData metricsData) {
...
     @Override
     public void sendMetricsDataToStorage(CollectRep.MetricsData metricsData) {
-        if (metricsDataProducer != null) {
-            ProducerRecord<Long, CollectRep.MetricsData> record =
-                    new ProducerRecord<>(kafka.getMetricsDataToStorageTopic(), metricsData);
-            metricsDataProducer.send(record);
-        } else {
-            log.error("metricsDataProducer is not enabled");
-        }
+        // The message is already sent via sendMetricsData since metricsDataToStorageConsumer
```
Calvin979: Hi, it must be a mistake. This method is not deprecated, metricsDataToStorageConsumer should subscribe to `kafka.getMetricsDataToStorageTopic()`

Calvin979 (changes requested):

hertzbeat-common/src/main/java/org/apache/hertzbeat/common/queue/impl/KafkaCommonDataQueue.java
starryCoder: Let me try to fix this error now.

hertzbeat-startup/src/main/resources/application.yml
```diff
@@ -150,6 +150,7 @@ common:
     kafka:
       servers: 127.0.0.1:9092
       metrics-data-topic: async-metrics-data
+      metrics-data-to-storage-topic: metrics-data-to-storage-topic
```
Duansg, line 153: Pls synchronize the changes to `/script/application.yml`

starryCoder: > Pls synchronize the changes to `/script/application.yml`


the changes have been synchronized.

Duansg (approved):

PR #3860, opened 2025-11-18
[improve] Added simple exponential backoff strategy

Duansg:
## What's changed?

Please refer to: [#3848](https://github.com/apache/hertzbeat/pull/3848)

For details:
1. Added an exponential backoff utility class
2. Added support for configuring the timeout for Redis brpop.
3. Added millisecond-level exponential backoff for all consumers blocking data acquisition.
4. Add new test cases.

## Checklist

## Add or update API


bigcyy:
Thank you very much for the PR and review invitation! I looked over the code, and it seems to differ slightly from my understanding. I have the following questions:

1.  I support changing `rpop` to `brpop` because it prevents CPU spin, but when `brpop` returns `null` after hitting a timeout due to its blocking nature, does the external fallback algorithm need to continue blocking for an even longer time? Can we just directly enter the next round of `brpop` or set a longer blocking time for `brpop`?
2.  My understanding is that the fallback strategy is intended for when `brpop` encounters an exception (e.g., a network connection failure). In such cases, we can attempt to reconnect. If the reconnection fails, we block and attempt to reconnect again, and only after reaching the retry limit do we terminate the consumer thread.

Duansg:
> 1. I support changing `rpop` to `brpop` because it prevents CPU spin, but when `brpop` returns `null` after hitting a timeout due to its blocking nature, does the external fallback algorithm need to continue blocking for an even longer time? Can we just directly enter the next round of `brpop` or set a longer blocking time for `brpop`?
> 2. My understanding is that the fallback strategy is intended for when `brpop` encounters an exception (e.g., a network connection failure). In such cases, we can attempt to reconnect. If the reconnection fails, we block and attempt to reconnect again, and only after reaching the retry limit do we terminate the consumer thread.

I have indeed considered this aspect. Currently, for processing logic like `Kafka`/`Redis`, both normal unresponsive data and exceptions return `Null`. Initially, I thought unified handling would suffice, but this does indeed delay entry into the next iteration when the queue is empty. I believe we should now distinguish between unknown exceptions and normal empty queues, implementing backoff strategies only when unknown exceptions occur.

PR #3875, opened 2025-11-29
chore: add .editorconfig for consistent coding style

yuluo-yx:


.editorconfig
```diff
@@ -0,0 +1,15 @@
+# http://editorconfig.org
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 2
```
tomsun28: hi, the indent size here is different from the java file indent.

yuluo-yx: should be 4?

tomsun28: yes, but the web-app's ts,js code, the indent_size is 2. Can this configuration be configured differently based on the directory?

yuluo-yx: got it, let me update it.

yuluo-yx: config rule refer: https://github.com/editorconfig/editorconfig/wiki/Projects-Using-EditorConfig

tomsun28 (approved):

PR #3922, opened 2025-12-21
[improve] optimize the function of creating monitoring using AI

LiuTianyou:
## What's changed?

1. optimize AI's monitoring creation capabilities and provide secure form entry for sensitive information.

image

image

2. support for protected mode by adding an extra prompt .
image

## Checklist

## Add or update API

### DIFF
--- a/hertzbeat-ai/src/main/java/org/apache/hertzbeat/ai/tools/impl/MonitorToolsImpl.java
+++ b/hertzbeat-ai/src/main/java/org/apache/hertzbeat/ai/tools/impl/MonitorToolsImpl.java
@@ -17,7 +17,6 @@
 
 package org.apache.hertzbeat.ai.tools.impl;
 
-import com.fasterxml.jackson.core.type.TypeReference;
 import com.usthe.sureness.subject.SubjectSum;
 import java.util.Optional;
 import lombok.extern.slf4j.Slf4j;
@@ -41,6 +40,7 @@ import org.springframework.ai.tool.annotation.ToolParam;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.apache.hertzbeat.common.entity.manager.Monitor;
 import org.apache.hertzbeat.common.entity.manager.Param;
+import tools.jackson.core.type.TypeReference;
 
 import java.util.ArrayList;
 import java.util.List;


hertzbeat-ai/src/main/resources/prompt/system-message.st
```diff
@@ -138,11 +218,88 @@ CRITICAL WORKFLOW Do all of this iteratively with user interaction at each step:
...
+     Monitor type: mysql
+     Public parameters collected:
+     - host: 192.168.1.10
+     - port: 3306
+     - name: Production Database
+     - interval: 60
+     Private parameters requiring secure collection:
+     - username (database username)
+     - password (database password)
+     Please complete configuration via the secure form."
```
tomsun28: Hi, although it looks like the user is being guided step by step to fill in the information, in reality each step is a brand-new LLM call. This unnecessarily consumes tokens and increases repeated context.

tomsun28: The suggestion is to shorten the steps. If the current context does not contain the required parameter values, the AI should ask the user for all parameters at once, and let the user provide them in a single input before proceeding. If the user’s prompt already includes some parameters, they can be added directly.

hertzbeat-ai/src/main/java/org/apache/hertzbeat/ai/tools/impl/MonitorToolsImpl.java
```diff
@@ -194,44 +210,45 @@ public String queryMonitors(
...
     public String addMonitor(
-            @ToolParam(description = "Monitor name (required)", required = true) String name,
-            @ToolParam(description = "Monitor type: website, mysql, postgresql, redis, linux, windows, etc.", required = true) String app,
-            @ToolParam(description = "Collection interval in seconds (default: 600)", required = false) Integer intervals,
-            @ToolParam(description = "Monitor-specific parameters as JSON string. "
-                    + "Use get_monitor_additional_params to see required fields. "
-                    + "Example: {\"host\":\"192.168.1.1\", \"port\":\"22\", \"username\":\"root\"}",
-                    required = true) String params,
-            @ToolParam(description = "Monitor description (optional)", required = false) String description) {
+        @ToolParam(description = "The id for current conversation", required = true) Long conversationId,
```
tomsun28: hi I found a new tool param `conversationId` was added. When we use it as MCP server in agent like claude code, how it can get the `conversationId`?

tomsun28:
Hi, i think whether it is provided to the built-in HertzBeat AI or exposed as an MCP tool for external agents, the tool’s capabilities should be atomic, with the AI model deciding how to orchestrate and invoke them. The design should not be driven by traditional user UI interactions, but by agent-centric thinking. The required data may be collected in multiple ways—for example, the AI may gather it from the user’s prompt, or an agent may read it in bulk from text and then add it.

hertzbeat-ai/src/main/java/org/apache/hertzbeat/ai/tools/impl/MonitorToolsImpl.java
LiuTianyou: Sorry, I overlooked the compatibility issue when the tool is called by MCP. I should provide an additional tool through enhancing the existing add_monitor tool, so that it can be invoked in scenarios where monitoring is added using the secure form method, such as add_monitor_protected. The conversationId parameter is only used in this tool.

LiuTianyou:
> Hi, i think whether it is provided to the built-in HertzBeat AI or exposed as an MCP tool for external agents, the tool’s capabilities should be atomic, with the AI model deciding how to orchestrate and invoke them. The design should not be driven by traditional user UI interactions, but by agent-centric thinking. The required data may be collected in multiple ways—for example, the AI may gather it from the user’s prompt, or an agent may read it in bulk from text and then add it.

I agree with your point that tools provided for AI calls should have higher versatility. I think we could offer some options to users, when they use ai assistant in hertzbeat， allowing them to choose the mode of AI involvement. For now, let’s refer to these as Protected Mode and Trusted Mode.

In Protected Mode, we could optimize the usage methods and data security based on general-purpose tools, enabling AI to integrate deeply into Hertzbeat.

In Trusted Mode, the process is entirely handed over to AI, allowing it to fully utilize the existing general-purpose tools to operate Hertzbeat.

tomsun28:
> > Hi, i think whether it is provided to the built-in HertzBeat AI or exposed as an MCP tool for external agents, the tool’s capabilities should be atomic, with the AI model deciding how to orchestrate and invoke them. The design should not be driven by traditional user UI interactions, but by agent-centric thinking. The required data may be collected in multiple ways—for example, the AI may gather it from the user’s prompt, or an agent may read it in bulk from text and then add it.
> 
> I agree with your point that tools provided for AI calls should have higher versatility. I think we could offer some options to users, when they use ai assistant in hertzbeat， allowing them to choose the mode of AI involvement. For now, let’s refer to these as Protected Mode and Trusted Mode.
> 
> In Protected Mode, we could optimize the usage methods and data security based on general-purpose tools, enabling AI to integrate deeply into Hertzbeat.
> 
> In Trusted Mode, the process is entirely handed over to AI, allowing it to fully utilize the existing general-purpose tools to operate Hertzbeat.

HI, the two modes seem like a good approach, would this correspond to two sets of Function Tools? Our current AI features are rapidly evolving, so it is recommended to use a single set of tools to achieve the desired consistency. 
Regarding security, the secret data will be transmitted to the LLM for input parameter filling, shouldn't we also pay attention to this, considering that other data is self-hosted?

LiuTianyou:
> HI, the two modes seem like a good approach, would this correspond to two sets of Function Tools? Our current AI features are rapidly evolving, so it is recommended to use a single set of tools to achieve the desired consistency. 
> Regarding security, the secret data will be transmitted to the LLM for input parameter filling, shouldn't we also pay attention to this, considering that other data is self-hosted?

Yes, only one toolset is used. The tools needed for the other mode are simply wrappers and enhancements on the existing tools.

Regarding security, in protected mode, private data is not sent to the LLM. After being collected via forms, parameters are populated when the AI ​​calls the `addMonitor` tool. In trusted mode, all data is sent to the LLM for processing. I think we should explain the differences between the two modes in the documentation and let users choose for themselves.

lynx009:
`com.fasterxml.jackson.core.type.TypeReference` needs to be replaced with `tools.jackson.core.type.TypeReference`
:
With the update of the project, JsonUtil's current TypeReference is tools.jacksoncore.type.TypeReference.
截屏2026-03-27 17 15 50
截屏2026-03-27 17 16 17

截屏2026-03-27 17 22 48

Issue #3961, opened 2026-01-12
[BUG] TDengine encountered a write exception in version 1.8.0


proyy:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

[BUG] 版本1.8.0 TDengine  DNS服务器监控 大量sql插入字段报错
2026-01-12T11:23:13.478987256+08:00 2026-01-12 11:23:13 [warehouse-persistent-data-storage] ERROR org.apache.hertzbeat.warehouse.store.history.tsdb.tdengine.TdEngineDataStorage - TDengine ERROR (216): sql: INSERT INTO `dns_additional_10_1_1_1_53` USING `dns_additional__super` TAGS (10.1.1.1:53) VALUES  (1768188193475, '{}', '{ . 0 CLASS1232 OPT [ ; payload 1232, xrcode 0, version 0, flags 0] }', NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL), desc: syntax error near '10.1.1.1:53) values  (1768188193475, '{}', '{ . 0 CLASS1232 OPT' (invalid data or symbol)
2026-01-12T11:35:13.227277225+08:00 2026-01-12 11:35:13 [warehouse-persistent-data-storage] ERROR org.apache.hertzbeat.warehouse.store.history.tsdb.tdengine.TdEngineDataStorage - TDengine ERROR (216): sql: INSERT INTO `dns_header_10_1_1_1_53` USING `dns_header__super` TAGS (10.1.1.1:53) VALUES  (1768188913224, '{}', 37.0, 'QUERY', 'NOERROR', 'qr aa rd ra ', 1.0, 4.0, 0.0, 1.0), desc: syntax error near '10.1.1.1:53) values  (1768188913224, '{}', 37.0, 'QUERY', 'NOER' (invalid data or symbol)

### Environment

```markdown
HertzBeat version(s): 1.8.0
```

Duansg:
tks! we will address this as soon as possible.

Saramanda9988:
Hi, This appears to be due to the lack of necessary escaping for special characters (such as dots, colons) in the instance during insertion in the #3863 refactoring

I'd be happy to take care of the fix to save you some time. Would you like me to submit a PR @Duansg

Duansg:
[...]

PR #3963, opened 2026-01-13
[Fix]: fix TAG type mismatch by introducing schema version for TDengine

Saramanda9988:
## What's changed?

Please refer to #3961 

This PR resolves the issue where the monitor TAG was incorrectly defined as BIGINT in v1.8.0, causing insertion failures for string-based instance
1. appended a _v2 suffix to both Super Table and Sub-table names (e.g., ..._super_v2). This bypasses TDengine's limitation where TAG column types cannot be modified after table creation.
2. Updated monitor TAG type to NCHAR(128) in the new v2 schema.
image
image

## Checklist

## Add or update API

### DIFF
--- a/hertzbeat-warehouse/src/main/java/org/apache/hertzbeat/warehouse/store/history/tsdb/tdengine/TdEngineDataStorage.java
+++ b/hertzbeat-warehouse/src/main/java/org/apache/hertzbeat/warehouse/store/history/tsdb/tdengine/TdEngineDataStorage.java
@@ -68,8 +68,14 @@ public class TdEngineDataStorage extends AbstractHistoryDataStorage {
     private static final String INSTANCE_NULL = "''";
     private static final String CONSTANTS_CREATE_DATABASE = "CREATE DATABASE IF NOT EXISTS %s";
     private static final String INSERT_TABLE_DATA_SQL = "INSERT INTO `%s` USING `%s` TAGS (%s) VALUES %s";
-    private static final String CREATE_SUPER_TABLE_SQL = "CREATE STABLE IF NOT EXISTS `%s` %s TAGS (monitor BIGINT)";
+    private static final String CREATE_SUPER_TABLE_SQL = "CREATE STABLE IF NOT EXISTS `%s` %s TAGS (instance NCHAR(128))";
     private static final String NO_SUPER_TABLE_ERROR = "Table does not exist";
+    /**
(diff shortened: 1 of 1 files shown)


Duansg (changes requested):
Doesn't Prometheus's task need to be handled?

Saramanda9988:
Hi @Duansg ,
﻿
Are you referring to the table name mismatch issue caused by users modifying the monitor name of Prometheus tasks?
If so, I find that this problem also exists in IotDB, QuestDB, and InfluxDB, but not in GreptimeDB (which only uses metrics) and VictoriaMetrics (which uses labels)
﻿
image
﻿
Perhaps we should address these related issues separately through a dedicated PR? This PR can focus on fixing TDengine's TAG issue.
﻿
Please let me know if I understood your concern correctly or you'd like me to fix TDengine's related issue just in this issue.

Duansg:
> Are you referring to the table name mismatch issue caused by users modifying the monitor name of Prometheus tasks?

That's correct, this is a historical issue, not caused by the instance reconstruction. I've reviewed the relevant logic and found it's not straightforward. Even a simple modification to ensure compatibility would require extensive regression testing on other dependent features. To ensure the issue remains traceable, I'll create a new PR to address this later.

Saramanda9988:
> > Are you referring to the table name mismatch issue caused by users modifying the monitor name of Prometheus tasks?
> 
> That's correct, this is a historical issue, not caused by the instance reconstruction. I've reviewed the relevant logic and found it's not straightforward. Even a simple modification to ensure compatibility would require extensive regression testing on other dependent features. To ensure the issue remains traceable, I'll create a new PR to address this later.

OK,thanks
If needed, I’m happy to help with some testing for that.

Duansg (approved):

Duansg:
> > > Are you referring to the table name mismatch issue caused by users modifying the monitor name of Prometheus tasks?
> > 
> > 
> > That's correct, this is a historical issue, not caused by the instance reconstruction. I've reviewed the relevant logic and found it's not straightforward. Even a simple modification to ensure compatibility would require extensive regression testing on other dependent features. To ensure the issue remains traceable, I'll create a new PR to address this later.
> 
> OK,thanks If needed, I’m happy to help with some testing for that.

👍🏿 ok, i suggest you reproduce the issue and file a ticket so we can track it  :)

Issue #3983, opened 2026-01-17
[BUG] Prometheus monitor history data query fails due to 'app' parameter inconsistency in some storages


Saramanda9988:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When querying historical metrics data for a Prometheus-type monitor, the query returns empty results even though the data has been successfully stored. Consequently, historical data charts display "No Data."

<img width="2559" height="1398" alt="Image" src="https://github.com/user-attachments/assets/d350f09c-5aa8-4e04-b989-46a7ea50f624" />

The issue stems from an inconsistency in the app parameter value between the storage and query phases for Prometheus monitors.

### Environment

```markdown
HertzBeat version(s):1.8.0
```

### Anything else?

At Storage (Monitor Creation/Modification): In MonitorServiceImpl.java:
```java
if (CommonConstants.PROMETHEUS.equals(monitor.getApp())) {
    appDefine.setApp(CommonConstants.PROMETHEUS_APP_PREFIX + monitor.getName());
}
```
For Prometheus monitors, AppDefine.app is set to prometheus_{monitorName} (e.g., prometheus_my_monitor). This value is used to construct table names in storage engines.

<img width="1218" height="390" alt="Image" src="https://github.com/user-attachments/assets/a518334d-47c0-4a69-8f78-44d8ece0dc7d" />

At Storage Execution (using TdEngine as an example):

- `String app = metricsData.getApp()` (retrieves the prefixed name from AppDefine).
- Table name becomes `app + "_" + metrics + "_" + instance + "_v2"`, resulting in prometheus_my_monitor_metrics_instance_v2.

<img width="2559" height="1527" alt="Image" src="https://github.com/user-attachments/assets/06fbe799-3b55-4592-bab2-6a60e51f0d55" />

At Query (Frontend to Controller):

- The frontend passes the app parameter from Monitor.getApp(), which is simply prometheus.
- MetricsDataController.java: Extracts `app = names[0]` (resulting in prometheus).

<img width="2559" height="1527" alt="Image" src="https://github.com/user-attachments/assets/30fab014-cdf6-4cd4-a71e-c3e593e66ce2" />

When the query reaches the storage layer (e.g., TdEngineDataStorage.java:360), getTable(app, metrics, instance) uses the raw prometheus string instead of the prefixed prometheus_my_monitor, failing to locate the correct table.

The following storages construct table names using the app parameter and are affected:
[...]

PR #3994, opened 2026-01-20
[refactor] Add monitorName parameter to history data query for consistent app handling

Saramanda9988:
## What's changed?

please refer to #3983 

In the method of querying historical data, add the `monitorName` parameter. When the app is prometheus, automatically concatenate the prefix _prometheus_{monitorName} to ensure that the query and storage use the same app value

- HistoryDataReader - add the monitorName parameter to the interface method (marked @ Deprecated)
- MetricsDataController - API path changed from/api/monitor/{instance}/metric/{metricFull} to/ -api/monitor/{instance}/metric/{monitorName}/{metricFull}
- MetricsDataService - add the monitorName parameter to the service method

### IotDb
a00ea231bfb24cfde6cd24f5306229c1

### Influxdb
5c84338dd3c5ff522b639242114431a4

### QuestDb
c9b31a715e736a313bb88396ba779e93

### TDengine
870f4df499ee3de29d6d2efb3f1f86ad

## Checklist

## Add or update API

### DIFF
--- a/hertzbeat-collector/hertzbeat-collector-basic/src/main/java/org/apache/hertzbeat/collector/collect/http/HttpCollectImpl.java
+++ b/hertzbeat-collector/hertzbeat-collector-basic/src/main/java/org/apache/hertzbeat/collector/collect/http/HttpCollectImpl.java
@@ -39,6 +39,7 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 import javax.net.ssl.SSLException;
+import javax.xml.XMLConstants;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.xpath.XPath;
@@ -106,6 +107,22 @@ import java.util.Collections;
  */
 @Slf4j
 public class HttpCollectImpl extends AbstractCollect {
+
+    /**
+     * Pre-compiled regex patterns for dangerous Xpath detection.
+     * Compiled once at class load for performance.
+     */
+    private static final List<Pattern> DANGEROUS_XPATH_PATTERNS;
+
+    static {
+        List<Pattern> patterns = new ArrayList<>();
+        for (String pattern : CollectorConstants.DANGEROUS_XPATH_PATTERNS) {
+            // Use CASE_INSENSITIVE to prevent bypass via case variations (e.g., //TEXT() vs //text())
+            patterns.add(Pattern.compile(pattern, Pattern.CASE_INSENSITIVE | Pattern.DOTALL));
+        }
+        DANGEROUS_XPATH_PATTERNS = Collections.unmodifiableList(patterns);
+    }
+
     private final Set<Integer> defaultSuccessStatusCodes = Set.of(
             HttpStatus.SC_OK,
             HttpStatus.SC_CREATED,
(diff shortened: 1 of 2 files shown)


web-app/src/app/routes/monitor/monitor-data-chart/monitor-data-chart.component.ts
```diff
@@ -264,7 +266,7 @@ export class MonitorDataChartComponent implements OnInit, OnDestroy {
     // load historical metrics data
     this.loading = `${this.i18nSvc.fanyi('monitor.detail.chart.data-loading')}`;
     let metricData$ = this.monitorSvc
-      .getMonitorMetricHistoryData(this.instance, this.app, this.metrics, this.metric, this.timePeriod, isInterval)
+      .getMonitorMetricHistoryData(this.instance, this.monitorName, this.app, this.metrics, this.metric, this.timePeriod, isInterval)
```
Duansg, line 269: Can we just directly convert the app here? That way, we won't need to adjust the REST API.

hertzbeat-ai/src/main/java/org/apache/hertzbeat/ai/tools/impl/MetricsToolsImpl.java
```diff
@@ -151,13 +151,14 @@ public String getRealtimeMetrics(
...
     public String getHistoricalMetrics(
-            @ToolParam(description = "Instance identifier (e.g., 'ip:port', 'ip', or 'domain')") String instance,
-            @ToolParam(description = "Monitor type (e.g., 'linux', 'mysql', 'http')", required = true) String app,
-            @ToolParam(description = "Metrics name (e.g., 'target', 'cpu', 'memory')", required = true) String metrics,
-            @ToolParam(description = "Field Parameter (e.g., 'usage', 'used', 'available')", required = false) String fieldParameter,
-            @ToolParam(description = "Label filter for specific instances", required = false) String label,
-            @ToolParam(description = "Time range (e.g., '1h', '6h', '24h', '7d')", required = false) String history,
-            @ToolParam(description = "Whether to aggregate data with intervals", required = false) Boolean interval) {
+        @ToolParam(description = "Instance identifier (e.g., 'ip:port', 'ip', or 'domain')") String instance,
+        @ToolParam(description = "Monitor name for querying historical metrics") String monitorName,
```
Duansg, line 155: Should monitorName be exposed to the Tool? I believe this is an unstable parameter and doesn't need to be exposed externally. It should remain within the internal normalization layer. I recommend performing app conversion internally.

web-app/src/app/routes/monitor/monitor-data-chart/monitor-data-chart.component.ts
Saramanda9988: hi,thanks for your cr.
Do you mean we should modify the app parameter directly in the fe? Maybe that would require adjusting greptimedb and vm to align with other tsdbs.
Just concerned this adds complexity to these two storages that we'll support long-term. What do you think?

Duansg: Thanks for your reply.

1. VictoriaMetricsClusterDataStorage And VictoriaMetricsDataStorage 

```
// saveData() method
if (metricsData.getApp().startsWith(CommonConstants.PROMETHEUS_APP_PREFIX)) {
    isPrometheusAuto = true;
}
```

```
// getHistoryMetricData()
String labelName = metrics + SPILT + metric;
if (CommonConstants.PROMETHEUS.equals(app)) { // app.startsWith(CommonConstants.PROMETHEUS_APP_PREFIX)
    labelName = metrics;
}
String timeSeriesSelector = Stream.of(
    LABEL_KEY_NAME + "=\"" + labelName + "\"",
    LABEL_KEY_INSTANCE + "=\"" + instance + "\"",
    CommonConstants.PROMETHEUS.equals(app) ? null : MONITOR_METRIC_KEY + "=\"" + metric + "\"" // app.startsWith(CommonConstants.PROMETHEUS_APP_PREFIX)
).filter(Objects::nonNull).collect(Collectors.joining(","));
```

2. GreptimeDbDataStorage: getHistoryMetricData() and getHistoryIntervalMetricData()

```
if (!CommonConstants.PROMETHEUS.equals(app)) { // default :  String timeSeriesSelector = LABEL_KEY_NAME + "=\"" + name + "\"" + "," + LABEL_KEY_INSTANCE + "=\"" + instance + "\"";
    timeSeriesSelector = timeSeriesSelector + "," + LABEL_KEY_FIELD + "=\"" + metric + "\"";
}
```

Theoretically, only the VM needs to be modified, right? And since the VM is saved through its mapping, maintaining consistency shouldn't be an issue. What do you think?

Saramanda9988: @Duansg , hi thanks for your reply

Yes, you're right. Let's do it this way.
I'll close this PR and submit a new one following your suggestion; it will be much simpler.

Issue #3955, opened 2026-01-09
[Task]Support Apache Doris as a time-series database for log.


zqr10159:
### Description

Requirements:
1.  Refer to https://doris.apache.org/docs/4.x/log-storage-.analysis and create tables based on the recommended content and Apache Hetzbeat's actual implementation.
2.  Implement an optional automatic log data expiration feature.
3. Implement identical log collection and query functionality to the existing Greptimedb, ensuring seamless user switching (module located in hertzbeat-log).
4. Log writing is recommended to use stream load mode; log queries are recommended to utilize database connection pools and MySQL JDBC.

Saramanda9988:
@zqr10159 Hi, I'd like to give it a try, could you please assign this issue to me? tks!

CrazyBeeline:
@Saramanda9988 Looking forward to your contribution

DanielCarter-stack:
Thanks for the proposal. Apache Doris as a log storage backend is indeed not currently implemented in HertzBeat—only GreptimeDB exists for this purpose.

The good news is that the warehouse module is well-architected for this extension:

- **Template available**: `hertzbeat-warehouse/src/main/java/org/apache/hertzbeat/warehouse/store/history/tsdb/greptime/GreptimeDbDataStorage.java` provides a complete reference implementation with `saveLogData()`, `queryLogsByMultipleConditions()`, `batchDeleteLogs()`, etc.
- **Interface contracts**: `HistoryDataReader.java` and `HistoryDataWriter.java` define the exact methods to implement
- **Data model**: `LogEntry.java` (OpenTelemetry standard) defines the log structure your table schema should map to

**Suggested implementation path:**
1. Create `DorisDataStorage.java` extending `AbstractHistoryDataStorage`
2. Use Doris Stream Load API for writes and MySQL JDBC for queries
3. Add `DorisProperties.java` for configuration (JDBC URL, Stream Load URL, credentials, TTL)
4. Implement TTL via Doris table properties or application-level cleanup

**A few clarifying questions:**
- Should TTL be enforced database-side (Doris table properties) or application-side (scheduled tasks)?
- Any requirements for Doris-specific features like rollup tables for log aggregation?
- Should both standalone and cluster Doris modes be supported?

This would be a valuable addition given the 12 existing TSDB implementations in the codebase prove the architecture's extensibility.

Issue #3954, opened 2026-01-09
[Task] Support Apache Doris as a time-series database for metrics.


zqr10159:
### Description

Requirements:
1.  Refer to https://doris.apache.org/docs/4.x/log-storage-analysis and create tables based on the recommended content and Apache Hetzbeat's actual implementation.
2.  Implement an optional automatic expiration feature for time-series data.
3. Implement identical functionality to existing solutions like Victoria Metrics and GrepTimeDB, ensuring seamless user switching(module located in hertzbeat-warehouse).
4. Recommend stream load for metric writing and database connection pooling with MySQL JDBC for metric querying.

Saramanda9988:
@zqr10159 Hi, I'd like to give it a try, could you please assign this issue to me? tks!

CrazyBeeline:
@Saramanda9988 Looking forward to your contribution

DanielCarter-stack:
Thanks for the feature request! HertzBeat currently supports 8 time-series databases in the `hertzbeat-warehouse` module (VictoriaMetrics, GreptimeDB, IoTDB, InfluxDB, TDengine, QuestDB, DuckDB) but lacks Apache Doris implementation. This is a valid feature gap.

**Current state:**
- `hertzbeat-warehouse/src/main/java/org/apache/hertzbeat/warehouse/store/history/tsdb/` has existing implementations that all extend `AbstractHistoryDataStorage` and implement `HistoryDataWriter`/`HistoryDataReader` interfaces
- Each backend uses `@ConditionalOnProperty` for conditional assembly (e.g., `warehouse.store.greptime.enabled`)
- `hertzbeat-warehouse/pom.xml` has no Doris-related dependencies

**Suggested approach:**
1. Create `doris/DorisDataStorage.java` (extend `AbstractHistoryDataStorage`, implement write via Stream Load, read via MySQL JDBC)
2. Create `doris/DorisProperties.java` for configuration (JDBC URL, Stream Load URL, credentials)
3. Add MySQL JDBC driver to `hertzbeat-warehouse/pom.xml`
4. Configure `warehouse.store.doris.enabled=true` in `application.yml`

**Questions before implementation:**
1. Minimum Doris version required (affects table DDL, partition/TTL, Stream Load compatibility)?
2. Should TTL use Doris native `data_tiering_ttl`/partition TTL or application-layer cleanup?

PR #4031, opened 2026-02-17
[feat] Support Apache Doris as time-series storage for both metrics and logs

Saramanda9988:
## What's changed?

please refer to #3954 #3955
This PR adds Apache Doris support as a time-series backend for both metrics and logs, aligned with HertzBeat’s existing storage abstraction and user-switching behavior.
- Added Doris-based storage implementation for metrics and logs in hertzbeat-warehouse.
- Added Doris table initialization logic for hzb_history and hzb_log based on Doris recommendations and HertzBeat schema needs.
- Added optional dynamic partition-based expiration for historical data/logs (retention/future window/time unit configurable).
- Added Stream Load write path for high-throughput ingestion.
- Kept JDBC write/query compatibility and integrated JDBC connection pooling for query/read path.
- Reused runtime worker pool (WarehouseWorkerPool) to avoid unmanaged raw threads.
- Added/updated Doris-related configuration entries in startup YAML files.

Using Apache Doris cluster with 1fe+3be for functional verification, not include long-duration soak/stability testing yet.

## screenshot
1ec326fdfb0a018f726f9bd3dde2c033
8e0fef494129b57251c82ef0e7cc60f2
833b1a3943a4f96c3f86306ed2f0c91d
b4ccd4aa7fc863a859f7e8d50c2d7b3c
a7d6d58961ed0e76acd7b034d09d3996
0a47337fbb6d279be7e4329a7e5314b8
d835ae3fa5931af487a126e0f4bb63b2

## Checklist

## Add or update API

### DIFF
--- a/home/blog/2023-12-11-hertzbeat-v1.4.3.md
+++ b/home/blog/2023-12-11-hertzbeat-v1.4.3.md
@@ -94,7 +94,7 @@ Detailed config refer to [Install HertzBeat via Docker](https://hertzbeat.apache
 * support victoriametrics as metrics data storage by @tomsun28 in [https://github.com/apache/hertzbeat/pull/1361](https://github.com/apache/hertzbeat/pull/1361)
 * Add time type to support query_time of mysql and mariadb by @Clownsw in [https://github.com/apache/hertzbeat/pull/1364](https://github.com/apache/hertzbeat/pull/1364)
 * add Clownsw as a contributor for code by @allcontributors in [https://github.com/apache/hertzbeat/pull/1365](https://github.com/apache/hertzbeat/pull/1365)
-* Error occured when I followed running steps to start Front-web by @Calvin979 in [https://github.com/apache/hertzbeat/pull/1366](https://github.com/apache/hertzbeat/pull/1366)
+* Error occurred when I followed running steps to start Front-web by @Calvin979 in [https://github.com/apache/hertzbeat/pull/1366](https://github.com/apache/hertzbeat/pull/1366)
 * add Calvin979 as a contributor for doc by @allcontributors in [https://github.com/apache/hertzbeat/pull/1367](https://github.com/apache/hertzbeat/pull/1367)
 * enriches the cncf landscape by @tomsun28 in [https://github.com/apache/hertzbeat/pull/1368](https://github.com/apache/hertzbeat/pull/1368)
 * Fix flaky test in CollectUtilTest by @bbelide2 in [https://github.com/apache/hertzbeat/pull/1371](https://github.com/apache/hertzbeat/pull/1371)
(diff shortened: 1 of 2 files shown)


hertzbeat-warehouse/src/main/java/org/apache/hertzbeat/warehouse/store/history/tsdb/doris/DorisDataStorage.java
```diff
@@ -0,0 +1,1250 @@
...
+        }
+    }
+
+    /**
+     * Build CREATE TABLE SQL statement
+     */
+    private String buildCreateTableSql() {
+        DorisProperties.TableConfig config = properties.tableConfig();
+
+        StringBuilder sql = new StringBuilder();
```
zqr10159: Consider using template strings instead of StringBuilder? The former is easier to maintain.

home/i18n/zh-cn/docusaurus-plugin-content-docs/current/start/doris-init.md
```diff
@@ -0,0 +1,160 @@
...
+你可以通过安装包或 Docker 部署 Doris。生产环境建议参考官方部署文档：
+
+- Doris 官方文档：[Quick Start](https://doris.apache.org/docs/4.x/gettingStarted/quick-start/)
+
+对 HertzBeat 接入来说，至少需要保证：
+
+- FE MySQL 服务端口可访问（默认 `9030`）
+- FE HTTP 服务端口可访问（默认 `8030`）
+
+### 前置检查
```
zqr10159, line 37: The final product of Hertzbeat will not include the MySQL-JDBC driver. Documentation must include instructions for MySQL-JDBC, referencing MySQL monitoring.

hertzbeat-startup/src/main/resources/application.yml
```diff
@@ -225,6 +225,38 @@ warehouse:
...
+        partition-future-days: 3
+        buckets: 8
+        replication-num: 1
+      pool-config:
+        minimum-idle: 5
+        maximum-pool-size: 20
+        connection-timeout: 30000
+      write-config:
+        # Write mode: jdbc (default, suitable for small/medium scale) or stream (high throughput)
+        write-mode: jdbc
```
zqr10159, line 246: It seems streamload has the highest priority in the code. If streamload is unavailable, it falls back to JDBC mode. So why is JDBC the default here? If users are unfamiliar with Doris's write capabilities, the default JDBC mode can impact performance during large-scale writes.

Saramanda9988: Hi, thanks for your cr, In my opinion,I choose jdbc as default because it has lower barrier to entry for testing and verification — users can start with official Docker image and verify connectivity immediately, while stream mode requires additional network config (HTTP port, redirect-policy) that may complicate initial setup; maybe we can encourage migration to stream for production via clearer documentation without making it the default.What do you think?

zqr10159: Good idea. We should at least have clearer documentation.

zqr10159:
I believe the direction of the modification is correct, but there are currently errors in the compilation and testing.

Saramanda9988:
> I believe the direction of the modification is correct, but there are currently errors in the compilation and testing.

I am running the test locally. I suspect that there may be a problem with the Doris related test case. I'll repair it

zqr10159:
lgtm

PR #4054, opened 2026-03-08
[bugfix] correct typo in sureness monitor delete verb definition

brettgervasoni:
## What's changed?

In the script/sureness.yml file there is a typo in the permission definitions for one of the DELETE calls. Instead of using three equals signs, two were used. This is invalid and defaults to the lower levelled user (a standard user), being able to call the DELETE verb when only an admin user is expected to be able to perform this operation.

By default, Sureness defaults to allow. This is a low severity security issue.

As a result, a non-admin user can use the DELETE verb to delete monitors.

I've rectified this by simply fixing the typo'd definition.

## Checklist

## Add or update API

### DIFF
--- a/script/sureness.yml
+++ b/script/sureness.yml
@@ -25,7 +25,7 @@ resourceRole:
   - /api/monitor/**===get===[admin,user,guest]
   - /api/monitor/**===post===[admin,user]
   - /api/monitor/**===put===[admin,user]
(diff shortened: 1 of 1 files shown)


zqr10159:
@brettgervasoni  Hello, I noticed there are still a few errors in sureness.yml that haven't been fixed. Could you help resolve the remaining issues? I'd be very grateful!

PR #4055, opened 2026-03-08
feat[common]: add protocl entity check

yuluo-yx:


hertzbeat-common-core/src/main/java/org/apache/hertzbeat/common/entity/job/protocol/PrometheusProtocol.java
```diff
@@ -107,8 +112,28 @@ public static class Authorization {
 
     @Override
     public boolean isInvalid() {
+        if (!validateIpDomain(host) || !validPort(port)) {
```
yuluo-yx, line 115: Other fields besides these should be optional configurations, therefore no validation is included here.

PR #4101, opened 2026-03-29
fix(manager): set instance to 'unknow' for dynamic monitors on modification

pentium100:
## What's changed?

Normalize monitor instances when adding or modifying monitors.

- dynamic service-discovery monitors consistently use `unknown`, preventing a
  null instance from reaching immutable metadata construction
- static monitor edits replace or remove an explicit old port according to the
  current port parameter
- raw IPv6 hosts are never mistaken for `host:port`; a configured port produces
  the unambiguous bracketed form (`[2001:db8::1]:443`)
- bracketed IPv6, IPv4, and domain instances remain stable across repeated edits
- the branch is merged with current `master` and the earlier formatting-only
  churn has been removed

## Regression proof

The latest-master baseline reproduced the service-discovery failure with
`expected: <unknown> but was: <memory>`. The corrected implementation passes:

- all 46 `MonitorServiceTest` tests
- 11 focused modify/instance tests, including domain, IPv4, raw IPv6, bracketed
  IPv6, port addition, replacement, and removal
- Maven Checkstyle and `git diff apache/master --check`

OCR delegation resolved the Java correctness, boundary, performance, and thread
safety rules. The test diff was manually reviewed because the deterministic
selector excludes test paths; no credible high or medium finding remains.

AI assistance: used for conflict resolution, regression coverage, and test
iteration.
Human validation: reproduced the old failure and ran the complete
`MonitorServiceTest` suite on Java 25.
Risk notes: an unbracketed IPv6 literal is always treated as a host; ports are
represented with brackets to avoid an ambiguous final numeric segment.


yuluo-yx:
hi @pentium100 pls see the copilot review comment.

zqr10159 (dismissed):

hertzbeat-manager/src/main/java/org/apache/hertzbeat/manager/service/impl/MonitorServiceImpl.java
```diff
@@ -391,13 +397,13 @@ public void modifyMonitor(Monitor monitor, List<Param> params, String collector,
         String portWithMark = (Objects.isNull(portParam) || !StringUtils.hasText(portParam.getParamValue()))
                 ? ""
                 : SignConstants.DOUBLE_MARK + portParam.getParamValue();
-        if (Objects.nonNull(instance)) {
+        if (Objects.nonNull(instance)&&!IpDomainUtil.isHasPortWithMark(instance)) {
```
Duansg: pls check for ci issues

Duansg (changes requested):

zqr10159 (dismissed):

hertzbeat-manager/src/main/java/org/apache/hertzbeat/manager/service/impl/MonitorServiceImpl.java
```diff
@@ -382,22 +382,28 @@ public void modifyMonitor(Monitor monitor, List<Param> params, String collector,
...
+            .filter(param -> PARAM_FIELD_PORT.equals(param.getField()))
+            .findFirst()
+            .orElse(null);
         String portWithMark = (Objects.isNull(portParam) || !StringUtils.hasText(portParam.getParamValue()))
-                ? ""
-                : SignConstants.DOUBLE_MARK + portParam.getParamValue();
-        if (Objects.nonNull(instance)) {
+            ? ""
+            : SignConstants.DOUBLE_MARK + portParam.getParamValue();
+        if (Objects.nonNull(instance) && !IpDomainUtil.isHasPortWithMark(instance)) {
```
Duansg: I have a question here: if this includes port-skipping concatenation, will changes made by the user to the port not take effect?

Duansg (changes requested):
In addition, please update the regression test cases for this change.

hertzbeat-manager/src/main/java/org/apache/hertzbeat/manager/service/impl/MonitorServiceImpl.java
pentium100: i have modified the code , check if instance contains port at first, if so  then remove it.

hertzbeat-manager/src/main/java/org/apache/hertzbeat/manager/service/impl/MonitorServiceImpl.java
```diff
@@ -382,22 +375,31 @@ public void modifyMonitor(Monitor monitor, List<Param> params, String collector,
...
-                .orElse(null);
+            .filter(param -> PARAM_FIELD_PORT.equals(param.getField()))
+            .findFirst()
+            .orElse(null);
         String portWithMark = (Objects.isNull(portParam) || !StringUtils.hasText(portParam.getParamValue()))
-                ? ""
-                : SignConstants.DOUBLE_MARK + portParam.getParamValue();
+            ? ""
+            : SignConstants.DOUBLE_MARK + portParam.getParamValue();
+        if (IpDomainUtil.isHasPortWithMark(instance)){
```
Duansg: I took a look at the code, and I have a question: if `monitor.setInstance("unknown")` is true, will this line of code execute?

Conversely, if that's not the case, I've noticed that the value of `instance` doesn't include port information during editing, so this line of code would never execute?

Please help confirm this issue.

Duansg (changes requested):

hertzbeat-manager/src/test/java/org/apache/hertzbeat/manager/service/MonitorServiceTest.java
```diff
@@ -624,14 +624,118 @@ void modifyMonitor() {
...
+        reset(monitorDao);
         Monitor existOkMonitor = Monitor.builder().jobId(1L).intervals(1).app("app").name("memory").instance("host")
                 .id(monitorId).build();
         when(monitorDao.findById(monitorId)).thenReturn(Optional.of(existOkMonitor));
         when(monitorDao.save(any(Monitor.class))).thenThrow(RuntimeException.class);
 
         assertThrows(MonitorDatabaseException.class,
                 () -> monitorService.modifyMonitor(dto.getMonitor(), dto.getParams(), null, null));
+
+        // 测试逻辑：当 scrape 不等于 static（且不为空），并且 instance 为空时，instance 应被设置为 "unknown"
```
Duansg: Please revise this into English.

Duansg (changes requested):

hertzbeat-manager/src/main/java/org/apache/hertzbeat/manager/service/impl/MonitorServiceImpl.java
pentium100: The instance is not edited directly on the frontend; instead, the host is assigned to the instance during submission, and the port parameter is appended on the backend to form the final instance. There are two scenarios:
1. When the scrape type is "static," the frontend host is mandatory. The frontend assigns the host to the instance, and the backend appends the port to form the `host:port` format, which is then assigned to the instance.
2. When the scrape type is "sd," the frontend host is empty, resulting in an empty instance. This is passed to the backend, but the modification logic does not handle this case, leading to an error at the following line:
```java
Map<String, String> metadata = Map.of(CommonConstants.LABEL_INSTANCE_NAME, monitor.getName(),
CommonConstants.LABEL_INSTANCE, monitor.getInstance());
```
Therefore, we need to add a check: if this is an "sd" type monitor, set the instance to "unknown." Given the current situation, it is true that the instance cannot contain a port. Are you suggesting that we should append the port directly without checking if the instance already contains one?

Duansg: Please thoroughly test all page functions before determining whether there is an issue here. And don’t just rely on AI assistance.

Duansg:
pls check ci

hertzbeat-manager/src/main/java/org/apache/hertzbeat/manager/service/impl/MonitorServiceImpl.java
```diff
@@ -90,6 +90,7 @@
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.Arrays;
```
Duansg, line 93: pls remove this import

Duansg (changes requested):

Duansg:
`org.apache.hertzbeat.manager.service.MonitorServiceTest#modifyMonitor` please review this method, as it may cause CI issues later on.

Duansg (changes requested):
@pentium100 Hi, Please continue reviewing the code. If you encounter any issues, please refer to the [Code Style and Quality Guide](https://hertzbeat.apache.org/docs/community/code_style_and_quality_guide) first.

zqr10159 (changes requested):
The null-instance fix for service-discovery monitors is directionally correct, but the new port-removal path corrupts valid IPv6 instances. `IpDomainUtil.isHasPortWithMark("2001:db8::1")` treats the final numeric segment as a port, so lines 394-395 truncate an unbracketed IPv6 address to `2001:db8:` even when no port parameter is supplied. Please use an IPv6-safe host/port representation or parser and add focused modify-monitor cases for IPv4/domain hosts plus bracketed and unbracketed IPv6, both with and without ports. The large unrelated indentation churn should also be removed before merge.

zqr10159 (approved):
Re-reviewed after the latest-master merge. The previous IPv6 truncation and formatting churn are gone; dynamic monitor identity is non-null, and explicit port replacement/removal is covered for domains, IPv4, raw IPv6, and bracketed IPv6. The complete MonitorServiceTest suite passes locally (46 tests) with Checkstyle.

PR #4116, opened 2026-04-10
add initial connection event for SSE emitter

jingshaoyi:
- Send CONNECTED event when creating SSE emitter to confirm connection
- Validate connection availability immediately on establishment
- Enable frontend to display connection status in real-time
- Clean up invalid emitters early if initial send fails

## Checklist

## Add or update API

### DIFF
--- a/hertzbeat-log/src/main/java/org/apache/hertzbeat/log/notice/LogSseManager.java
+++ b/hertzbeat-log/src/main/java/org/apache/hertzbeat/log/notice/LogSseManager.java
@@ -96,6 +96,16 @@ public class LogSseManager {
         emitter.onError((ex) -> removeEmitter(clientId));
 
         emitters.put(clientId, new SseSubscriber(emitter, filters));
+        try {
+            String initEventId = "init-" + System.currentTimeMillis();
+            emitter.send(SseEmitter.event()
+                .id(initEventId)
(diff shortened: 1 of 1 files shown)


jingshaoyi (commented):
1

bigcyy:
Thanks for the contribution.

  I don't think this PR is needed for the current log SSE scenario.

  The frontend already uses `EventSource.onopen` / `onerror` to maintain the connection status, so sending an extra
  `CONNECTED` event from the backend does not add much value here. Also, the frontend currently only listens for
  `LOG_EVENT`; this new `CONNECTED` event is not consumed by the log stream page.

  There is also a lifecycle concern: this `emitter.send(CONNECTED...)` is called inside `createEmitter()`, before the
  controller returns the `SseEmitter` to Spring MVC. At that point Spring may not have initialized the underlying
  response handler yet, so the event can be queued as an early send attempt rather than actually written immediately. If
  the initial event fails later during emitter initialization, the `catch` block in `createEmitter()` may not catch it,
  and cleanup may not happen as expected.

  For these reasons, I suggest not adding this initial `CONNECTED` event. If we need a stronger business-level
  subscription acknowledgement in the future, we should design it together with frontend consumption and proper emitter
  lifecycle/error handling.

Issue #3737, opened 2025-09-02
[Task] Add database account expiry monitoring metrics


delei:
### Description

In the database-type monitoring templates, add monitoring metrics for database account expiry and support alerting functionality.

### What needs to be done

1. Add or modify monitoring metrics within the template.
2. Ensure monitoring data functions correctly.
3. Supports alert configuration.
4. Modify the template documentation and add new metric descriptions.
5. Please provide test cases if necessary.

### Additional context

- This task list is related to the issue #2757 
- Please refer to the sample provided in the PR #3674 

### Task List

- [x] `app-mysql.yml` (@BhanuNidumolu)
- [x] `app-mongodb.yml`
- [ ] `app-mariadb.yml`
- [ ] `app-oceanbase.yml`
- [ ] `app-dm.yml`
- [ ] `app-greenplum.yml`
- [ ] `app-kingbase.yml`
- [ ] `app-opengauss.yml`
- [ ] `app-postgresql.yml` (@zxuexingzhijie )
- [ ] `app-sqlserver.yml`
- [ ] `app-tidb.yml`
- [ ] `app-vastbase.yml`

delei:
If the community accepts this task list, it is expected that these tasks will be completed within the next milestone period.

Please feel free to add any missing tasks, and feel free to comment or submit a PR.

>😃 Welcome new contributors and as their first contribution.

tomsun28:
hi welcome! 👍

XL-Zhao-23:
I'm interested in working on this task. Could you please assign it to me?

delei:
> I'm interested in working on this task. Could you please assign it to me?

Hi, @XL-Zhao-23 

Welcome. If you are interested in contributing, please indicate which specific task you have in mind. Community members may assign it to you later.

BhanuNidumolu:
Hi, I’d like to start by implementing database account expiry metrics for MySQL (app-mysql.yml) and extend it to other databases after review.

zxuexingzhijie:
Hi, I’d like to start by implementing database account expiry metrics for pg (app-postgresql.yml) @tomsun28 ,plta

delei:
> Hi, I’d like to start by implementing database account expiry metrics for MySQL (app-mysql.yml) and extend it to other databases after review.

Welcome.

delei:
> Hi, I’d like to start by implementing database account expiry metrics for pg (app-postgresql.yml) @tomsun28 ,plta

Welcome.

[...]

PR #4131, opened 2026-04-29
[feat]:add db account monitoring metrics for postgresql.

wy471x:
Related [Task#3737](https://github.com/apache/hertzbeat/issues/3737):

All changes:
1.Add monitoring info to postgresql yaml.
2.Update description markdown docs.

Test case:
截屏2026-04-30 10 13 36
截屏2026-04-30 10 13 14


wy471x:
@Duansg Hello, I have fixed the CI issue. Please take a look when you have time.

hertzbeat-manager/src/main/resources/define/app-postgresql.yml
```diff
@@ -1076,3 +1076,94 @@ metrics:
...
+          CASE
+            WHEN valuntil IS NULL THEN NULL
+            ELSE EXTRACT(DAY FROM (valuntil - NOW()))::INTEGER
+          END as days_left,
+          CASE
+            WHEN valuntil IS NULL THEN 'OPEN'
+            WHEN valuntil < NOW() THEN 'EXPIRED'
+            ELSE 'OPEN'
+          END as account_status
+        FROM pg_user
```
Duansg: Starting with version 8.1, PostgreSQL adopted the `roles` model. I see that the old `users` compatibility view is being used here. Should we consider switching to `roles`?

Duansg (changes requested):

hertzbeat-manager/src/main/resources/define/app-postgresql.yml
wy471x: Okay, I also agree with your opinion. I have made modifications to this. Please take a look when you have time. Thank you.

hertzbeat-manager/src/main/resources/define/app-postgresql.yml
```diff
@@ -1076,3 +1076,95 @@ metrics:
...
+      queryType: multiRow
+      sql: |
+        SELECT
+          rolname AS usename,
+          oid AS usesysid,
+          CASE WHEN rolcreatedb THEN 'Y' ELSE 'N' END as usecreatedb,
+          CASE WHEN rolsuper THEN 'Y' ELSE 'N' END as usesuper,
+          rolvaliduntil AS valuntil,
+          CASE
+            WHEN rolvaliduntil IS NULL THEN NULL
```
Duansg, line 1148: I have found that PostgreSQL's special time semantics, such as `infinity`, are not currently supported. If this configuration is present, it may cause an exception to be thrown.

Duansg (changes requested):

hertzbeat-manager/src/main/resources/define/app-postgresql.yml
wy471x: I have already handled the special cases of the time type. Please take a look when you have time.

Duansg (approved):

PR #4154, opened 2026-06-11
fix: avoid blocking modify monitor by redundant detect call

### LINKED ISSUE #3812: [BUG] It got stuck for a long time when modifying the monitoring save settings, and eventually timed out, but it was confirmed that the save was successful in the end.

### Current Behavior

It got stuck for a long time when modifying the monitoring save settings, and eventually timed out, but it was confirmed that the save was successful in the end.

The monitoring service is operating normally and can also obtain monitoring values properly.

As shown in the figure below, I am modifying the monitoring Intervals.

Image

Image

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
HertzBeat version(s):1.7.3
docker
```

### Debug logs

_No response_

### Anything else?

_No response_

wilmerdooley:
## What's changed?

This makes `modifyMonitor` in `MonitorServiceImpl` no longer block on a synchronous detection pass. Previously the update path called `detectMonitor` inline, which runs a synchronous collection on the request thread and could block for a long time, eventually timing out the API even though the save itself succeeded.

Removing that synchronous detect on its own would let the final full-row save persist the status carried on the request, overwriting the live status that the real-time collection path maintains (raised in review). So the change also preserves the existing status before the save, `monitor.setStatus(preMonitor.getStatus())`, on the common path that both the active and paused branches reach. The collector continues to refresh the status on its next scheduled run.

A unit test (`testModifyMonitorPreservesLiveStatus`) covers both that a live monitor keeps its status across a modify and that a paused monitor is not flipped.

Resolves #3812

## Checklist

- [ ] I have written the necessary doc or comment.

## Add or update API

### DIFF
--- a/hertzbeat-manager/src/main/java/org/apache/hertzbeat/manager/service/impl/MonitorServiceImpl.java
+++ b/hertzbeat-manager/src/main/java/org/apache/hertzbeat/manager/service/impl/MonitorServiceImpl.java
@@ -431,12 +431,6 @@ public class MonitorServiceImpl implements MonitorService {
                 newJobId = collectJobScheduling.updateAsyncCollectJob(appDefine, collector);
             }
             monitor.setJobId(newJobId);
-
-            // execute only in non paused status
-            try {
-                detectMonitor(monitor, params, collector);
-            } catch (Exception ignored) {
-            }
         }
 
         // After the update is successfully released, refresh the database
@@ -451,6 +445,8 @@ public class MonitorServiceImpl implements MonitorService {
             // force update gmtUpdate time, due the case: monitor not change, param change.
             // we also think monitor change
             monitor.setGmtUpdate(LocalDateTime.now());
+            // preserve the live status owned by the real-time collection path; the modify request must not overwrite it
+            monitor.setStatus(preMonitor.getStatus());
             // update or open grafana dashboard
             if (monitor.getApp().equals(CommonConstants.PROMETHEUS) && grafanaDashboard != null) {
                 if (grafanaDashboard.isEnabled()) {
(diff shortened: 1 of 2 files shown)


hertzbeat-manager/src/main/java/org/apache/hertzbeat/manager/service/impl/MonitorServiceImpl.java
```diff
@@ -431,12 +431,6 @@ public void modifyMonitor(Monitor monitor, List<Param> params, String collector,
                 newJobId = collectJobScheduling.updateAsyncCollectJob(appDefine, collector);
             }
             monitor.setJobId(newJobId);
-
```
Duansg, line 436: This method currently updates the monitor's status in real time; removing it directly would cause status overwriting issues.

Duansg (changes requested):

hertzbeat-manager/src/main/java/org/apache/hertzbeat/manager/service/impl/MonitorServiceImpl.java
wilmerdooley: We resolved this by setting the status of the updated monitor object to the existing preMonitor status using `monitor.setStatus(preMonitor.getStatus())` right before saving. This ensures that the real-time status updates are not overwritten during modifications. A corresponding unit test `testModifyMonitorPreservesLiveStatus` has also been added to verify this behavior.

wilmerdooley:
Thanks again for the catch. One note on the behavior tradeoff with this approach: dropping the synchronous detect means the status now reflects the last value the real-time collection path wrote and reconverges on the next collect cycle, rather than being re-measured on the spot at edit time. If keeping that instant post-edit refresh matters here, I would be glad to move the detect off the request thread and run it asynchronously instead, which would still keep this preserve line so the save does not race it. Otherwise the preserve-only fix as pushed keeps the status correct, and I am happy to take the async direction if it fits the project better.

Duansg:
> Otherwise the preserve-only fix as pushed keeps the status correct, and I am happy to take the async direction if it fits the project better.

You're welcome, i don't think asynchronous processing is necessary in this situation; we can simply wait for the next cycle after making the changes.

Duansg (approved):

wilmerdooley:
Thanks again for the review and approval, @Duansg. I noticed the branch is now a little behind master. I am happy to rebase or update it if that helps this land, whatever you prefer. No rush.

PR #4163, opened 2026-06-18
[web]bugfix: clear request body when http method changes to GET

### LINKED ISSUE #2298: [BUG] When http method change by POST to GET,the request body didn't delete

### Current Behavior

This will cause the monitoring to return an HTTP 415 error
unless you chose POST again and the request body is manually deleted

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
HertzBeat version(s):v1.6.0
```

### Debug logs

_No response_

### Anything else?

_No response_

Wyh-max-star:
### What changed

Clear the request body when the HTTP method is changed from POST to GET in the HTTP/API monitor form.

### Why

Fixes #2298.

Previously, when users changed the HTTP method from POST to GET, the previously entered request body could still remain in the form and be submitted with the GET configuration.

### How tested

* Added/updated the related monitor form unit test.
* Manually verified that changing the HTTP method from POST to GET clears the request body.
* Verified that switching back to POST still allows editing the request body.


Duansg:
@Wyh-max-star 👍🏿 thanks for your pr, please check copilot reply.

PR #4167, opened 2026-06-20
[collector]bugfix: honor XML response charset

### LINKED ISSUE #2852: [BUG] HttpCollectImpl XML parsing assumes UTF-8

### Current Behavior

https://github.com/apache/hertzbeat/blob/1fe70a60cc82d1fa83b9d2af3b46b047480796dc/hertzbeat-collector/hertzbeat-collector-basic/src/main/java/org/apache/hertzbeat/collector/collect/http/HttpCollectImpl.java#L251

If you have a String, you don't need to convert to byte array (which is almost a waste of memory).

DocumentBuilder has a parse(InputSource) method.
https://docs.oracle.com/javase/8/docs/api/javax/xml/parsers/DocumentBuilder.html#parse-org.xml.sax.InputSource-

InputSources can be constructed to wrap StringWriters that wrap the String.

### Expected Behavior

Don't convert Strings to byte arrays unnecessarily wasting memory and causing parse issues. Imagine if the XML has an XML declaration that has an encoding that is not UTF-8. If you already have the String, the parser will ignore the value. If you convert to a byte array, the parser will use the XML encoding value but you have explicitly converted to UTF-8 in your code so these encodings may not match.

### Steps To Reproduce

_No response_

### Environment

```markdown
HertzBeat version(s): latest
```

### Debug logs

_No response_

### Anything else?

_No response_

hutiefang76:
## What's changed

Close #2852.

- Decode XML HTTP responses with the charset from `Content-Type` when present.
- Fall back to the XML declaration charset before using UTF-8.
- Keep non-XML response parsing behavior unchanged.
- Add regression coverage for ISO-8859-1 XML without an HTTP charset and unsupported XML charset fallback.

## Verification

- `JAVA_HOME=$(/usr/libexec/java_home -v 25 2>/dev/null || /usr/libexec/java_home) ./mvnw -pl hertzbeat-collector/hertzbeat-collector-basic -am -Dtest=HttpCollectImplTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `JAVA_HOME=$(/usr/libexec/java_home -v 25 2>/dev/null || /usr/libexec/java_home) ./mvnw -pl hertzbeat-collector/hertzbeat-collector-basic -am test`
- `git diff --check`


tomsun28 (approved):
👍

PR #4220, opened 2026-07-20
[improve] mask secret fields of notice receiver in rest api response

Duansg:
## What's changed

The notice receiver GET endpoints returned credentials in clear text. This masks them in the api response while keeping the configuration editable.

- Mask the 11 credential fields of `NoticeReceiver` in the three receiver GET endpoints.
- Restore the stored value on `editReceiver` / `sendTestMsg`, so saving without retyping a credential keeps the existing one.
- A credential of 12+ chars keeps its last 4 visible (`******voJM`), shorter ones are fully masked. The placeholder is fixed-length so it does not leak the real length.

Masking is applied at the controller only — `getReceiverById` still returns real values because `AlertNoticeDispatch` needs them to send notifications. No frontend change required.

## Tests

Tests have been conducted on existing and new data, and they were successfully passed.

## Checklist

## Add or update API

Jul-20-2026 20-23-02


tomsun28 (approved):
LGTM!

PR #4239, opened 2026-07-24
fix(web-app): point threshold user guide link to correct docs

orangeCatDeveloper:
### What's changed

The **User Guide** link on the Threshold Rules page (`/alert/setting`) opened
the app's own 404 page instead of the documentation.

The help component bound its link to the i18n key `alert.help.center.setting`,
which is not defined in any locale. The i18n pipe returns the key unchanged, so
the link rendered `href="alert.help.center.setting"` and the router resolved it
to `/exception/404`. The correct key `alert.help.setting.link` already exists in
every locale and points at the alarm threshold docs; this PR simply references
it.

One-line change in `alert-setting.component.html`.

### Screenshots

**Before** (User Guide → app 404):

before-404

**After** (User Guide → alarm threshold docs):
after-docs

### Test plan

- Set the UI language to any locale, open **Threshold** page, click **User Guide**.
- Before: opens the in-app 404 page.
- After: opens `https://hertzbeat.apache.org/docs/help/alert_threshold`.


tomsun28 (approved):
lgtm

PR #4242, opened 2026-07-24
improve(dashboard): hide empty category cards, add empty state

### LINKED ISSUE #3087: [Question] On the homepage, can a div be removed when there is no related app available? When it's empty, display a div similar to "Click to add a monitor".
### Question

首页是否可以在未存在相关APP时，移除div。当全空时展示一个类似“点击添加一个监控”的div
On the homepage, can a div be removed when there is no related app available? When it's empty, display a div similar to "Click to add a monitor".

目前我仅有操作系统monitor，但是首页却在轮询滚动其他，感觉不直观
Currently, I only have the operating system monitor, but the homepage is polling and scrolling others, which feels non-intuitive.

Image

orangeCatDeveloper:
### What & why

The dashboard showed an auto-scrolling carousel of all monitor categories even
when a category had `0` monitors, and a brand-new user with no monitors just saw
a carousel of empty cards. Closes #3087.

### Design decision

- Show only the categories that actually have monitors; drop the empty ones.
- With no monitors at all, replace the carousel with a welcoming empty state
  (illustration + an **Add Monitor** button) rather than scrolling empty cards —
  kept compact and styled to match the direction discussed in #3087.

Existing chart-click / navigation behavior is unchanged.

### Screenshots

**No monitors — before / after**

before-A-empty-carousel

after-A-empty-state

**A few monitors — before / after**

before-B-all-categories

after-B-only-nonempty

### Test plan

- 0 monitors -> empty state with Add Monitor button (no carousel of empty cards).
- Monitors in 2 categories -> only those 2 cards show, no auto-scroll.
- Monitors in 5+ categories -> carousel returns as before.


tomsun28 (approved):
Great! thanks.

PR #4252, opened 2026-07-26
[docker-compose]feature: require configurable password via .env instead of hardcoded default (#4218)

### LINKED ISSUE #4218: [Feature] Maybe the password in docker-compose.yaml can be random rather than 123456?
### Feature Request

Hi, I had installed hertzbeat with docker compose, ref to :https://hertzbeat.apache.org/zh-cn/docs/start/docker-compose-deploy

But I found that the PostgreSQL username and password used by Hertzbeat are root/123456. That doesn't seem like a good idea in a production environment.  After installed hertzbeat, I used db tools to connect my server with root/123456, and everything in hertzbeat shows.

Maybe something can be done? A random password, or using .env, or anything else.

Thanks

_No response_

_No response_

_No response_

### Additional context

_No response_

Bhavya-Sonigra:
Fixes #4218

## What's changed?

The `hertzbeat-postgresql-victoria-metrics` docker-compose deployment previously hardcoded PostgreSQL credentials (`root`/`123456`) directly in `docker-compose.yaml`, making it trivial for anyone with network access to connect to the database using well-known default credentials.

This PR:
- Introduces a `.env.example` file with placeholder credentials and instructions to copy it to `.env`
- Updates `docker-compose.yaml` to read `POSTGRES_USER`/`POSTGRES_PASSWORD` from environment variables instead of hardcoding them
- Uses Docker Compose's `${VAR:?error message}` syntax so `docker compose up` fails with a clear error if `POSTGRES_PASSWORD` isn't set, rather than silently falling back to a known-weak default
- Passes the same credentials to the `hertzbeat` service via `SPRING_DATASOURCE_USERNAME`/`SPRING_DATASOURCE_PASSWORD` so the app connects correctly
- Adds `.env` to `.gitignore` so real passwords are never committed

Tested locally: verified the stack starts and connects correctly with a password set in `.env`, and verified `docker compose up` fails with a clear error message when `.env`/`POSTGRES_PASSWORD` is missing.

This fix covers the `hertzbeat-postgresql-victoria-metrics` variant specifically, matching the deploy guide referenced in the issue. Happy to follow up with the same fix for the other docker-compose variants if maintainers want that as a separate PR.

## Checklist
- [x] I have read the Contributing Guide
- [ ] I have written the necessary doc or comment.
- [ ] I have added the necessary unit tests and all cases have passed.


Bhavya-Sonigra:
Hi @tomsun28, thanks for taking a look at this PR. I’ve resolved the merge conflicts and updated the branch with the latest master. I noticed the PR is currently waiting for an approving review and workflow approval. Is there anything else needed from my side before it can move forward?

Bhavya-Sonigra:
Closing this PR since the changes were incorporated into #4257 and are already present in master

PR #4258, opened 2026-07-28
[fix] bind greptime compose data-store ports to localhost

Aias00:
## Summary
Security hardening for the `hertzbeat-postgresql-greptimedb` quick-start compose bundle.

The Postgres and GreptimeDB data-store ports were mapped to `0.0.0.0` (all interfaces). Both services run with **default credentials shipped in the compose file**:
- `POSTGRES_PASSWORD: 123456`
- greptime static user provider `greptime=greptime`

Binding those ports to `0.0.0.0` lets any host that can reach the machine authenticate to the underlying data stores using only those published defaults.

## Fix
Bind the host-side port mappings to `127.0.0.1`. The `hertzbeat` container reaches both data stores over the internal compose network (`hertzbeat` network, service DNS names `postgresql` / `greptime`) and does not need these host mappings at all — they exist only for the operator to inspect the data stores from the host, so localhost binding preserves that use case.

- `127.0.0.1:15432:5432` (postgres)
- `127.0.0.1:14000-14003:4000-4003` (greptime)

The hertzbeat web/collector ports (`1157`/`1158`) are intentionally left unchanged.

## Out of scope (noted, not addressed here)
The default credentials themselves remain in the bundle for quick-start ergonomics. Operators should still override `POSTGRES_PASSWORD` / the greptime user provider in production. Binding to localhost is the robust mitigation against remote use of those defaults.

### DIFF
--- a/script/docker-compose/hertzbeat-postgresql-greptimedb/docker-compose.yaml
+++ b/script/docker-compose/hertzbeat-postgresql-greptimedb/docker-compose.yaml
@@ -32,8 +32,11 @@ services:
       timeout: 5s
       retries: 5
       start_period: 30s
+    # Bind data-store ports to localhost only: these services carry default
+    # credentials (see POSTGRES_PASSWORD / the greptime static user provider)
+    # and are only meant for the operator on the host, not for remote access.
     ports:
-      - '15432:5432'
+      - '127.0.0.1:15432:5432'
     environment:
       POSTGRES_USER: root
       POSTGRES_PASSWORD: 123456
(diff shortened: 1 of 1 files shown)


yuluo-yx (approved):

zqr10159:
Thanks for narrowing the data-store exposure. This is useful partial hardening, but it does not clear the full quick-start finding yet: the HertzBeat service remains remotely published while the bundle still documents/ships a predictable administrator credential path. Please either require an explicit first-run/override credential, or document and enforce that the manager is not remotely exposed until initialization. I would treat this PR as the database-port portion only.

zqr10159:
I added the narrow dependent follow-up #4295 for the remaining published listeners. This PR correctly scopes the Postgres and GreptimeDB ports, but the quick-start still publishes HertzBeat 1157/1158 on every interface while the bundled manager credential is unchanged. #4295 makes those listeners resolve to 127.0.0.1 by default and requires an explicit HERTZBEAT_BIND_ADDRESS choice for remote deployment.

For the default quick-start exposure described here, I would treat #4258 plus #4295 as the complete mitigation. The static credentials still need replacement before an operator deliberately enables remote access.

PR #4267, opened 2026-07-29
[fix] fix align cors configuration with header based authentication

Duansg:
## What's changed?

Cross-origin access itself is unchanged: origins stay open and preflight requests are still answered, so no working request is affected.

Added a unit test that drives the real filter and asserts both halves — credentials are not allowed, and the preflight still succeeds.

## Checklist

## Add or update API


zqr10159 (approved):
The CORS configuration now matches HertzBeat's header-based authentication model: cross-origin API access remains available, while ambient credentials are no longer permitted. The regression tests cover both halves of that contract, and the current-head checks pass.

PR #4272, opened 2026-07-29
[fix] require an account to subscribe to the alert and manager streams

Duansg:
## What's changed?

Require authentication for the alert/manager SSE streams; the front end switches to fetch so it can carry credentials, plus a connection timeout and a concurrency cap.

## Checklist

## Add or update API


zqr10159 (dismissed):
Reviewed against the 1.9.0 pre-release alert and manager SSE findings. The streams are removed from the authentication exclusions, the browser sends bearer authorization, and emitter lifetime/capacity are bounded. The focused local contracts pass (10 tests) and CI is green.

Aias00 (approved):
## Review: require auth for the alert/manager SSE streams ✅ APPROVED

Solid security hardening with a real root cause, consistent config, and thorough tests. Approving.

### Why this matters (the vulnerability it closes)
Previously `/api/alert/sse/**` and `/api/manager/sse/**` sat in `excludedResource`. Sureness evaluates the exclusion tree **before** any credential check, so an anonymous `curl -N` could subscribe and receive every alert the deployment raised — internal hostnames, addresses, metric values and full alert content — because `AlertNoticeDispatch` broadcasts each alert to *every* subscriber with no per-subscriber filtering. This PR removes the two paths from `excludedResource` and adds them to `resourceRole` as `===get===[admin,user,guest]`. The new `SurenessSseRuleTest` pins exactly that contract (paths require an account, are no longer anonymous, and the log-stream scoping is unchanged).

### Correctness checks
- **Config is consistent**: the same `-4/+2` change is applied to all six `sureness.yml` files (startup, all 5 docker-compose variants, `script/`), so no deployment path is left exposing the streams anonymously.
- **Backend lifecycle is genuinely hardened** by the new shared `SseEmitterRegistry` (replacing the two divergent per-stream copies):
  - finite `EMITTER_TIMEOUT_MILLIS` (30 min) instead of `Long.MAX_VALUE` — a client that vanishes without closing no longer holds a request thread until restart;
  - `maxEmitters = 1000` concurrency cap;
  - the cap is enforced by a `compareAndSet` slot claim (`heldSlots`), which actually holds under a parallel burst (verified by `testCeilingHoldsWhenSubscriptionsArriveTogether` with 200 racers);
  - `onError` now releases the slot — this fixes a pre-existing bug where one stream leaked slots of errored subscribers until process restart. The class comment documents this exactly.
- **Frontend carries credentials correctly**: `AuthorizedSseService` reads the token per attempt via `fetch` with `Authorization: Bearer …` (browser `EventSource` cannot send that header, which is *why* the endpoints were anonymous before). It also owns reconnect backoff, `AbortController` cleanup, and token refresh on 401/403 (the `fetch` path bypasses `HttpClient`, so the interceptor never sees the 401 and refresh must happen here). `authorizedHeaders()` reads from localStorage each attempt so a refreshed token is picked up immediately.
- **Interceptor tweak** (default.interceptor.ts) is a defensible companion: a token-refresh failure on a `SILENT_HTTP_ERROR` request no longer force-redirects to login — consistent with the SSE service doing its own refresh.

### Tests
Good coverage: `SurenessSseRuleTest` (authz contract), `SseEmitterRegistryTest` (finite timeout, cap, slot reclaim on error/complete failure, race), `AlertSseManagerTest`/`ManagerSseManagerTest` (event-name delivery, registry-bounded), plus `default.interceptor.spec.ts`, `auth.service.spec.ts`, `authorized-sse.service.spec.ts`.

### Non-blocking suggestions
1. The role is `[admin,user,guest]`. If a deployment wants the streams only for real users, dropping `guest` is a one-line tightening — fine either way, just flagging the intent.
2. `maxEmitters` is `@Setter` but I don't see a config binding wiring it from `application.yml`; it currently relies on the `1000` default. Consider exposing it (e.g. `hertzbeat.sse.max-subscriptions`) so operators can tune per deployment.
3. Nothing here, but worth a follow-up: the same anonymous-exposure pattern should be audited for any other endpoint still in `excludedResource` that carries tenant data.

Bottom line: this fixes a genuine information-disclosure bug without breaking the reconnect/refresh behaviour the UI depends on. Approved.

PR #4273, opened 2026-07-29
[fix] bound what one anonymous push request can consume

Duansg:
## What's changed?

Add three configurable caps to the push gateway: auto-created monitors, request body size, and sample count.

## Checklist

## Add or update API


zqr10159:
There is still a concurrency hole in the auto-created-monitor cap. `containsKey`/`size` is checked before `computeIfAbsent`; multiple new job/instance keys can all observe capacity and then insert, exceeding the configured maximum. The current tests exercise sequential calls only. Please make reservation atomic (for example with a semaphore/explicit reservation or a truly bounded cache) and add a concurrent burst test that proves the map never exceeds the cap. The request-body and sample-count limits are still useful independently.

zqr10159 (approved):
The previously reported capacity race is resolved. New monitor slots are reserved atomically before persistence, failed creations release their reservation, concurrent pushes for the same pair share one creation, and the pair key no longer has separator collisions. The regression suite now covers concurrent distinct pairs, same-pair deduplication, failed saves, stale misses, startup accounting, and parser-side body/sample bounds. Code review is approved; the new-head Backend Build and Maven E2E required checks are still running and must pass before merge.

PR #4275, opened 2026-07-29
[fix] validate sql reaching the alert query executor

Duansg:
## What changed?

- Validate every SQL statement at the selected executor so alternate alert-expression spellings cannot bypass the read-only boundary.
- Fail closed on unsupported SQL while preserving validated Greptime RANGE, ALIGN, and FILL syntax.
- Keep raw log queries restricted to `hertzbeat_logs`; periodic metric SQL can use dynamic tables only in the configured database and cannot use schema-qualified relations.
- Bound alert previews and periodic evaluations to 8,192 query characters, one day of lookback, 1,000 returned rows, and the existing warehouse HTTP timeouts.
- Restrict alert-definition reads to `admin,user`; query preview, create, update, and import are admin-only in the packaged policy and every deployment copy.

## Validation

- `./mvnw -pl hertzbeat-startup -am -Dtest=SqlSecurityValidatorTest,DataSourceServiceTest,SurenessResourceRuleTest test -DskipITs -Dsurefire.failIfNoSpecifiedTests=false -DfailIfNoTests=false`
  - 55 SQL validator tests passed
  - 49 datasource service tests passed
  - 15 packaged/deployment authorization tests passed
- `./mvnw -pl hertzbeat-startup -am -DskipTests package`
  - all 24 selected modules packaged successfully
- Checkstyle and `git diff --check` passed.

## Checklist

- [x] I have read the Contributing Guide.
- [x] I have added the necessary comments and regression tests.
- [x] Focused tests and source packaging pass.

Human validation: reproduced each authorization, budget, and table-scope regression before implementation; reran the complete focused suites and source package on Java 25.
Risk notes: alert-query preview and recurring query authoring now require an administrator; unsupported vendor SQL fails closed except for the explicitly normalized Greptime range clauses.


zqr10159:
This is useful executor-level read-only validation, but it does not close the complete alert-query exposure. Low-privilege callers can still issue broad telemetry reads, expensive SELECTs, and raw PromQL; SELECT-only parsing also permits unions/subqueries across available tables. Please add an authorization/ownership boundary for preview and authoring, scope readable sources/tables, and enforce time/range/cost limits for both SQL and PromQL. I would treat the current patch as one defense-in-depth layer, not a complete remediation.

zqr10159 (approved):
Reproduced the remaining authorization, resource-budget, and cross-schema gaps after the latest-master merge. The final branch closes those paths, preserves validated Greptime range syntax, and passes 55 SQL validator, 49 datasource, 15 policy/deployment tests plus the 24-module source package. No blocking issue remains from my review.

PR #4276, opened 2026-07-29
[fix] stop serving the openapi document to anonymous callers

Duansg:
## What's changed?

The OpenAPI/Swagger document endpoints are no longer open to anonymous callers; scoped to admin.

## Checklist

## Add or update API


Duansg:
/swagger-ui/index.html is still served anonymously by the existing /**/*.html===get exclusion, but the page fetches /v3/api-docs/swagger-config and /v3/api-docs without an Authorization header — swagger-ui's Authorize button only applies to try-it-out calls, not to the spec fetch. The UI will therefore show "Failed to load API definition" for everyone, including admins.

The document itself remains reachable for an admin with a token:

curl -H "Authorization: Bearer $JWT" http://localhost:1157/v3/api-docs

If the community would rather not ship a UI page that cannot load, the follow-up is to set springdoc.api-docs.enabled / springdoc.swagger-ui.enabled to false by default and let deployments opt in; that is a product decision kept out of this PR.

zqr10159 (changes requested):
Restricting the generated OpenAPI document to administrators is the right security boundary, but the current change leaves `/swagger-ui/index.html` publicly reachable while the UI cannot attach an administrator token to its initial `/v3/api-docs/swagger-config` and `/v3/api-docs` requests. The result is a shipped page that shows “Failed to load API definition” for anonymous users and administrators alike. Please either disable the Swagger UI by default (with an explicit opt-in deployment path), or provide an authenticated loading flow that lets an administrator use it. Keep the document endpoints admin-only in either case.

Duansg:
**What changed**

- **Swagger UI authenticates itself.** `AuthorizedSwaggerIndexTransformer` replaces the springdoc `SwaggerIndexTransformer` bean and appends a `requestInterceptor` to `swagger-initializer.js` that attaches the stored HertzBeat token to same-origin requests. springdoc's built-in `swagger-ui.csrf.use-local-storage` cannot be used here: it writes the stored value verbatim, while HertzBeat keeps the bare JWT and callers add the `Bearer ` prefix. If the initializer ever ships in an unrecognised shape the transformer throws rather than serving an unauthenticated page.
- **One endpoint was missed.** springdoc serves grouped documents at `/v3/api-docs.yaml/{group}`; added `- /v3/api-docs.yaml/**===get===[admin]` to all copies. The old assertion targeted `/v3/api-docs/default.yaml`, a path springdoc does not serve.
- **All 9 in-tree `sureness.yml` copies are in sync**.
- **Enabling only `swagger-ui.enabled` no longer breaks startup.** The transformer is built from springdoc beans gated by `api-docs.enabled`, so the bean condition now matches both switches.
- Docs updated in English and zh-cn.

**Verified**

Signed in as an administrator on the same origin, `/v3/api-docs/swagger-config` and `/v3/api-docs` are requested with `Authorization: Bearer …`, both `200`, and the UI renders with no prompt. Without that session sureness answers `401` with a Digest challenge; administrator credentials get through, an authenticated non-`admin` account gets `403`. Unit tests cover the rules, the switches, and the transformer; checkstyle clean.


Aug-13-2026 01-58-02

zqr10159 (approved):
The previous Swagger UI regression is resolved. OpenAPI and Swagger UI are now disabled by default across the packaged and deployment configurations; when explicitly enabled, the UI attaches the existing HertzBeat JWT only to same-origin requests, while the document endpoints remain admin-only. The configuration-copy contracts, transformer tests, focused security tests, and current-head CI are all green.

PR #4277, opened 2026-07-29
[fix] restrict the actuator endpoints to admin

Duansg:
## What's changed?

Add role rules for /actuator/**, restricting them to admin.

## Checklist

## Add or update API


zqr10159:
Several compose/script Sureness files add `/actuator/**===get===[admin]` inside `excludedResource`. A resource listed there can bypass authentication before `resourceRole` is applied, so those distributions may remain exposed. Please move the rule into `resourceRole` consistently, assert that actuator paths are absent from every exclusion list, and add 401/403/admin-success tests across the packaged configurations. This PR also does not protect the collector management/Prometheus endpoint; that endpoint needs a separate bind/auth control.

zqr10159 (approved):

PR #4283, opened 2026-07-30
maintenance: canonicalize alert label fingerprints

zqr10159:
## Summary

- serialize calculator cache label maps as sorted, length-prefixed key/value pairs before hashing
- preserve label association while remaining deterministic across map iteration order
- document and test the compatibility boundary between this transient cache coordinate and the durable `SingleAlert.fingerprint`

## Compatibility evidence

`AlertUtil.calculateFingerprint` is used only for the in-memory calculator cache. `AlarmCacheManager` reconstructs that coordinate from persisted labels during startup. The durable alert identity is generated separately by `AlarmCommonReduce.generateAlertFingerprint`; this pull request does not change that path.

The added restart regression loads a persisted firing alert, rebuilds the new cache coordinate from its labels, resolves it through the cache, and verifies that its durable fingerprint remains unchanged. A separate reducer contract verifies the durable fingerprint format and its exclusion of timestamp labels.

Consequently, this change does not rewrite database identities or break group, silence, or inhibition references, and a versioned database migration or dual-read path is not required.

## Validation

- collision regression distinguishes `{a:b, c:d}` from `{a:d, c:b}` and map-order regression preserves deterministic output
- `./mvnw -pl hertzbeat-alerter -am -Dtest=AlertUtilTest,AlarmCacheManagerTest,AlarmCommonReduceTest,MetricsPeriodicAlertCalculatorTest,MetricsRealTimeAlertCalculatorTest,DbAlertStoreHandlerImplTest test -DskipITs -Dsurefire.failIfNoSpecifiedTests=false -DfailIfNoTests=false` — 22 tests passed
- `./mvnw -pl hertzbeat-alerter test -DskipITs` — 439 tests passed, 2 skipped
- Checkstyle and both Git diff whitespace checks passed
- no hard-coded CJK scan passed

Human validation: focused cache/reducer contracts and the complete alerter test suite passed after rebuilding the current reactor dependencies.
Risk notes: only the transient calculator cache coordinate changes. Persisted alert identity remains unchanged and the cache is rebuilt from labels at process startup.


zqr10159:
Author remediation update:

The upgrade blocker was removed by narrowing canonicalization to transient in-memory cache identity. Durable alert fingerprints and all database and external relationships keep their existing algorithm and values, so an alert firing before upgrade continues the same active record and can resolve after restart. No database, silence, acknowledgement, or inhibition migration is required.

Restart and firing-to-resolve regressions passed (22 focused tests), and the full alerter reactor passed (439 tests). All current GitHub checks are green. The author-side upgrade blocker is resolved; maintainer review is still required.

Duansg (approved):

PR #4288, opened 2026-07-30
maintenance: scope Grafana request authentication

zqr10159:
## Summary

- retain the existing `Authorization: Basic ...` header on every Grafana service-account request
- stop adding a duplicate, persistent credential interceptor to the shared HTTP client
- add a contract proving the local Basic header remains present while the shared client is not mutated

The three request paths still call `createHeaders()`. That unchanged helper builds the Basic header from the configured Grafana username and password. This change only removes the additional shared interceptor.

## Validation

`./mvnw -pl hertzbeat-grafana -Dtest=ServiceAccountServiceTest,DashboardServiceTest test -DskipITs -Dsurefire.failIfNoSpecifiedTests=false -DfailIfNoTests=false`

Live smoke: Grafana 13.1.1 with a real `RestTemplate` and Basic admin credentials successfully listed service accounts, created the HertzBeat service account, and issued its token. The temporary live test passed 1/1 and was not committed.

Human validation: focused Grafana service tests passed (4 tests); live Grafana 13.1.1 service-account flow passed (1 test); Checkstyle and diff checks passed.
Risk notes: request Authorization headers are unchanged; the persistent shared-client interceptor is no longer installed.


zqr10159:
Author validation update:

Basic authentication was not removed. Every Grafana request still uses createHeaders(), which builds the Authorization Basic header from the configured username and password. The removed code was only a second persistent BasicAuthenticationInterceptor on the shared RestTemplate, which accumulated credentials across requests.

This was verified against a real Grafana 13.1.1 instance using a real RestTemplate: listing service accounts, creating the HertzBeat service account, and issuing its token all succeeded. Focused tests passed (4 plus one temporary live smoke), all current GitHub checks are green, and the temporary Grafana test and runtime artifacts were not committed.

Duansg (approved):

PR #4291, opened 2026-07-30
maintenance: derive rule-scoped notice groups

zqr10159:
## Summary

- Route each grouped notice through the alerts that actually matched the selected rule.
- Recompute group labels and key, common labels, common annotations, status, fingerprint list, and audit time range from that scoped subset.
- Keep the stored group unchanged for database, plugin, and SSE consumers; label-free rules continue to forward the original group.
- Build independent payloads when multiple rules target the same receiver with different templates.

## Regression proof

The new outcome tests were first run against the previous PR head. They failed because the notice payload copied the original group status, grouping fields, common fields, and timestamps after filtering its alert list. A same-receiver/multiple-rule case also reused data from alerts outside each rule. The same contracts pass with this revision.

## Validation

- `./mvnw -pl hertzbeat-alerter -am -Dtest=AlertNoticeDispatchTest,NoticeConfigServiceTest test -DskipITs -Dsurefire.failIfNoSpecifiedTests=false -DfailIfNoTests=false` — 39 tests passed.
- `./mvnw -pl hertzbeat-alerter -am test -DskipITs -Dsurefire.failIfNoSpecifiedTests=false -DfailIfNoTests=false` — reactor passed; common-core 584, common-spring 102, warehouse 59, alerter 437 (2 skipped).
- `git diff --check` — passed.
- Touched files contain no hard-coded CJK text.

## Behavior and compatibility

Only label-filtered rule notices change. Their visible and derived fields now consistently describe the selected alerts. The persisted source group is not mutated. When single-alert audit times are unavailable, the scoped group retains the source group time as a compatibility fallback.

Group-label keys remain the original grouping dimensions, with values recomputed from the selected alerts. Routing remains group-then-scope rather than a larger route-then-regroup redesign.

Human validation: regression contracts failed on the previous head; focused and full Maven reactor tests passed as listed above.
Risk notes: templates that accidentally depended on common fields from non-matching alerts will now receive only rule-scoped data.


zqr10159:
Author remediation update:

Each label-filtered rule now receives a fully self-consistent derived group. groupLabels, groupKey, commonLabels, commonAnnotations, status, fingerprints, and time range are recomputed from that rule subset. The stored source group is not mutated, and multiple rules sharing one receiver receive independent payloads and templates.

Mixed firing and resolved, different annotation, same-receiver and multiple-rule, and source-immutability regressions passed (39 focused tests); the full alerter reactor passed (437 tests, 2 skipped). GitHub checks are currently queued. Maintainer review remains required.

zqr10159:
CI follow-up: backend build, Maven E2E, image E2E, license, and label checks have all completed successfully on the current head.

Duansg (approved):

PR #4293, opened 2026-07-30
maintenance: extend SFTP connection options

zqr10159:
## Summary

- Pin SFTP server identity to one or more configured SHA-256 host-key fingerprints.
- Accept a current and next fingerprint together so operators can rotate host keys without an outage.
- Use one protocol validation contract for monitor validation and collector pre-check.
- Reevaluate advanced form dependencies when the SFTP boolean changes, including persisted string values.
- Bound repeated warnings for the explicit unsafe opt-out and document trusted fingerprint acquisition, upgrade, and rotation in English and Chinese.

## Upgrade behavior

This is an intentional fail-closed change for 1.9.0. Existing SFTP monitors are **not** automatically migrated to skip verification. Before or immediately after upgrading, operators must either:

1. obtain and verify the server fingerprint through a trusted channel, then configure one or more `SHA256:...` values; or
2. explicitly select the dangerous skip-verification option as a short-lived recovery measure.

Until one of those policies is configured, the affected SFTP monitor reports a configuration error and does not connect. Plain FTP monitors are unchanged.

`ssh-keyscan` is documented only as a way to retrieve a key, not as proof of identity. Operators must compare the resulting fingerprint through a trusted channel.

## Regression proof

- A malformed skip-verification value previously passed collector pre-check when a valid fingerprint was present; it is now rejected by the shared protocol contract.
- Boolean changes previously did not reevaluate dependent advanced fields, so the fingerprint field could remain hidden; the monitor form now updates visibility and clears hidden values correctly.
- Persisted `"false"` strings are normalized as disabled rather than treated as truthy.
- Protocol serialization does not expose validation helpers as accidental JSON properties.

## Validation

- `FtpProtocolTest`: 10 passed.
- `FtpCollectImplTest`: 8 passed.
- Focused Angular monitor-form tests: 5 passed.
- Scoped frontend ESLint: passed with zero warnings.
- `./mvnw -pl hertzbeat-startup -am -DskipTests package`: all 24 reactor modules passed.
- Angular production build: passed (with the repository's existing initial-bundle budget warning).
- Checkstyle and `git diff --check`: passed.

Human validation: failing-first protocol and UI contracts, focused backend/frontend tests, Checkstyle, ESLint, source package proof, and production frontend build all passed.
Risk notes: existing SFTP monitors without an explicit policy fail closed after upgrade; the upgrade guide documents the required operator action and temporary recovery option.


zqr10159:
Author remediation update:

The rollout path now preserves existing SFTP monitors through an explicit one-time compatibility migration instead of making them fail immediately. New configurations still fail closed unless they provide a trusted fingerprint or explicitly select the dangerous compatibility option. Multiple fingerprints are accepted for current and next key rotation, unsafe connections emit controlled warnings, and the UI plus English and Chinese docs explain trusted acquisition, upgrade, rotation, and exit steps.

Protocol, verifier, and migration tests passed (15), monitor YAML parsing passed, and the 24-module startup source package passed. GitHub checks are currently queued. Maintainer review remains required.

zqr10159:
CI follow-up: the only failing check after the remediation is DOC CI Dead Link Check. It was rerun once and failed again on the same unchanged historical URL, https://gitee.com/hertzbeat/hertzbeat, returning HTTP 405. The changed SFTP documentation passed Markdown and filename checks and adds no failing external link. Backend build, Maven E2E, image E2E, license, and label checks all passed. I am leaving the unrelated global link configuration and historical pages out of this single-purpose PR; the DOC CI result should be retried after the external endpoint or repository-wide link policy is addressed.

zqr10159:
DOC CI has now passed on the unchanged current head (`6a368854b0`). The clean rerun completed Markdown, filename, dead-link, dependency-install, and documentation-build steps successfully. The earlier Gitee 405 was transient; no unrelated link-check policy change was needed. Together with the already green backend build, Maven E2E, image E2E, license, and label checks, this removes the remaining CI blocker.

Duansg:
Thanks — pinning host keys here makes sense. SshClient.setUpDefaultClient() inherits ClientBuilder.DEFAULT_SERVER_KEY_VERIFIER, which is AcceptAllServerKeyVerifier.INSTANCE (verified in sshd-core 2.13.1), so today any server key is accepted. A few points on the implementation.

The compatibility migration isn't one-time

SftpHostKeyCompatibilityMigration is a CommandLineRunner with no version record or completion marker, so it re-scans all ftp monitors on every startup and writes insecureSkipVerify=true into any SFTP monitor lacking both hostKeyFingerprint and insecureSkipVerify.

That's correct for the upgrade moment, but it keeps applying afterwards: a monitor that reaches the database without a host-key policy — via config import, a direct API call, or SQL — gets converted to "skip verification" at the next restart. The normal UI path is likely unaffected, since insecureSkipVerify has defaultValue: false and produces a param row, but the import/API paths bypass that.

Suggestion: make it a Flyway Java migration, or persist a completion flag so the backfill runs once. New monitors should opt into the unsafe mode explicitly rather than inherit it.

"Required" is declared three different ways

In app-ftp.yml, hostKeyFingerprint is required: false and hide: true, its display name says "（必填）/ Required", and the collector treats it as mandatory in both FtpProtocol.isInvalid() and preCheck().

hide: true puts it in the collapsed Advanced Parameters panel (monitor-form.component.html:122-131) rather than hiding it outright, but a user who doesn't expand that panel gets a rejected job referencing a field they never saw. Worth making the three agree — either mark it required and surface it, or drop "必填" from the label.

Pin the digest explicitly

KeyUtils.getFingerPrint(PublicKey) resolves through DEFAULT_DIGEST_HOLDER, an AtomicReference that's mutable via setDefaultFingerPrintFactory(...) and the KEY_FINGERPRINT_FACTORY_PROP system property. Since the docs and placeholder promise SHA256:, the explicit overload would be safer:

KeyUtils.getFingerPrint(BuiltinDigests.sha256, serverKey)

Fingerprint comparison is format-sensitive

Raw byte equality against the trimmed input means a fingerprint pasted without the SHA256: prefix, or with different base64 casing, fails with the generic "did not match" warning and no hint that the format was the issue. Normalizing, or rejecting malformed entries up front with a clear message, would save operators some guessing.

Minor: MessageDigest.isEqual is unnecessary here — fingerprints are public, there's no secret to leak through timing.

zqr10159:
Addressed the review findings in commit `a920268d2b`.

Changes:
- removed the unconditional startup migration that could silently opt future policy-less SFTP rows into `insecureSkipVerify`;
- validate configured fingerprints as `SHA256:base64`;
- pass Apache SSHD's SHA-256 digest factory explicitly rather than relying on its mutable global default;
- aligned the conditional UI wording and operator docs with the fingerprint-or-explicit-dangerous-skip contract;
- documented that HertzBeat never enables skip-verification automatically.

Human validation:
- `FtpCollectImplTest,FtpProtocolTest` passed across common-core and collector-basic, with manager compilation included;
- the digest regression changes SSHD's global default to MD5 and still proves SHA-256 verification;
- `git diff --check`.

Risk notes: operators choosing the explicit skip-verification option still accept MITM risk, and the collector logs that state.

Duansg:
The underlying issue is real — SshClient.setUpDefaultClient() uses AcceptAllServerKeyVerifier.INSTANCE, so SFTP collection currently accepts any host key. Scoping the change to ssl=true so plain FTP is untouched is right, and supporting two fingerprints for a rotation window plus documenting that ssh-keyscan alone doesn't authenticate a server are both good calls.

That said, I don't think this can go in as-is.

1. The upgrade path described in the PR body isn't in the diff, and existing monitors will break.

The description says:

> Existing ftp monitors with SFTP enabled and neither a fingerprint nor an explicit verification policy receive insecureSkipVerify=true once, before scheduler initialization. … SftpHostKeyCompatibilityMigrationTest

There's no migration class in the 7 changed files, and the docs actually state the opposite ("HertzBeat does not enable that option automatically"). So the code and docs agree on fail-closed; the description is stale.

I traced what actually happens to a pre-upgrade monitor. In CollectUtil.replaceSpecialCharacterIfNeeded():

Configmap param = configmap.get(replaceField);
if (param == null) { continue; }   // placeholder is left in place verbatim

Existing monitors have no rows for the new params, so:
- hostKeyFingerprint resolves to the literal "^_^hostKeyFingerprint^_^" — non-blank, so Assert.hasText passes
- Boolean.parseBoolean("^_^insecureSkipVerify^_^") is false
- Assert.isTrue(hasValidHostKeyFingerprints()) throws
- MetricsCollect:201 catches it and sets Code.FAIL

Every existing SFTP monitor goes unavailable right after upgrade, with the message "host key fingerprints must use the SHA256:base64 format" — for a field the operator never filled in. Either land the migration for real (plus a home/docs/start/upgrade.md note), or make the first release warn-only and enforce in the next one.

2. depend: ssl: [true] has no effect in the UI.

monitor-form.component.html:110-115 only routes radio parents to onDependChanged; boolean parents go to onParamBooleanChanged, which contains no depend handling at all (just the api/ftp ssl↔port defaulting). Every existing depend: in the repo is keyed on a radio field (snmpVersion, httpMethod, __sd_authType__) — a boolean parent is new here.

Combined with hide: true, both params land in the advanced list (monitor-new.component.ts:125) where Param.display defaults to true and the section renders on display !== false. Net effect: both fields are always visible, so plain-FTP users see a "DANGER: skip SFTP host key verification" checkbox. The "conditional on SFTP" claim isn't achieved.

Simplest fix is to drop the depend block and say "SFTP only" in the param name/tip, unless you want to add boolean-parent support to the form first.

3. FtpProtocol.isInvalid() is dead code.

There are only six .isInvalid() call sites in main (consul_sd, nacos_sd, nginx, registry, pop3, dns); FTP isn't one of them — FtpCollectImpl enforces via preCheck(). So the +48 lines in FtpProtocol and +48 in FtpProtocolTest cover a method that never runs in production, while the six lines in preCheck do the real work. The two have already diverged: isInvalid() rejects a malformed insecureSkipVerify value and rejects a bad fingerprint even when skip is on; preCheck does neither. Having preCheck delegate to isInvalid() would give one source of truth and make the tests cover the live path.

Smaller things:

- createServerKeyVerifier runs on every collection inside handleSftpCollect, so with insecureSkipVerify=true the warning fires once per interval per monitor. Worth deduplicating, or warn once and debug afterwards.
- The SHA256:[A-Za-z0-9+/]{43}=? pattern depends on KeyUtils.getFingerPrint emitting the SHA256: prefix, even though BuiltinDigests.sha256.getAlgorithm() is SHA-256 and MINA strips the hyphen internally — a quiet break waiting for a MINA upgrade. KeyUtils.checkFingerPrint(String, PublicKey) handles both the SHA256:base64 and MD5:xx:xx: forms and would remove the regex and the manual comparison.
- limit: 2048 is inert here: ParamDefine.limit is a Short documented as max 255, and only type: textarea-limit reads it (extra.textarea_limit) — plain textarea ignores it. (paramValue is varchar(8126), so capacity itself is fine.)
- serverKeyVerifierAlwaysUsesSha256 mutates JVM-global state via KeyUtils.setDefaultFingerPrintFactory. It's restored in finally, but it will interfere with other tests if surefire ever runs them in parallel in the same JVM.
- verifyServerKey(null, null, key) passes a null session and address, coupling the test to the fact that the current implementation ignores them.

Happy to re-review once the upgrade path and the depend wiring are sorted.

Duansg:
The compatibility migration described in the body is not in the diff — existing SFTP monitors will stop collecting on upgrade

The Summary and "Upgrade behavior" both promise a one-time migration setting insecureSkipVerify=true on legacy SFTP monitors, and Validation lists SftpHostKeyCompatibilityMigrationTest. The current head contains only 7 files (FtpCollectImpl, FtpProtocol and their tests, app-ftp.yml, two ftp.md docs) — no migration class, no SQL, no such test. Searching the diff for migration|compatibility|CommandLineRunner|Flyway|.sql returns nothing. It looks like the migration was dropped in a rebase and the description didn't follow.

Failure chain: a legacy monitor's params contain neither field → CollectUtil.replaceSmilingPlaceholder takes the else { value = null; break; } path, so both arrive as null → in preCheck, parseBoolean(ssl) is true and parseBoolean(null) is false, entering the new branch where Assert.hasText(null) throws → MetricsCollect:201-213 sets Code.FAIL → basic is the priority: 0 availability metric, so the monitor goes unavailable. Plain FTP is unaffected.

Please pick one and make the description match:

- Restore the migration. Note the earlier revision's concern still applies: a CommandLineRunner with no version or completion marker re-runs on every startup, so a monitor later created via import, the API, or direct SQL without a host-key policy gets silently flagged insecureSkipVerify=true — reversing the fail-closed guarantee this PR is built on. A Flyway migration or a persisted completion marker avoids that.
- Accept fail-closed. Better posture, and it removes the hole above — but Summary, "Upgrade behavior" and Validation all need rewriting, and the release notes must state that after upgrade every SFTP monitor stops collecting until an operator supplies a fingerprint verified through a trusted channel, or ticks the temporary skip option.

Validation should also be re-run against the actual head — SftpHostKeyCompatibilityMigrationTest doesn't exist there, so "15 tests passed" isn't reproducible.

Minor: depend: ssl on the two new params won't take effect. depend is only re-evaluated for radio fields (boolean dispatches to onParamBooleanChanged, which never touches depend), and ssl is boolean — so both fields will show unconditionally. Every existing depend: in the repo targets a radio field.

The verifier itself looks right: explicit BuiltinDigests.sha256, enforced SHA256:base64 format, multiple fingerprints for rotation, MessageDigest.isEqual comparison.

Aias00 (dismissed):
## Review: maintenance: extend SFTP connection options

**Verdict: ✅ APPROVED** — correct, fail-closed host-key pinning with rotation support.

### What this PR does
Hardens SFTP collector security:
- Pins the SFTP server identity to one or more SHA-256 host-key fingerprints.
- Supports a current + next fingerprint together, so operators can rotate host keys with no outage.
- Migrates legacy monitors (no host-key policy) to a temporary compatibility flag so they keep running, with controlled warnings.
- Marks the "skip verification" UI option as dangerous/conditional and documents trusted-fingerprint acquisition, upgrade, and rotation (EN + ZH).

### Assessment (verified the verification is real, not a no-op)
- **Fail-closed**: `preCheck` requires `hostKeyFingerprint` (and validates the `SHA256:base64` format) whenever verification is not explicitly disabled.
- **Correct crypto**: the verifier computes `KeyUtils.getFingerPrint(BuiltinDigests.sha256, serverKey)` — explicitly SHA-256, not the mutable default digest (a test proves it still uses SHA-256 even when the default factory is temporarily set to MD5). Match uses `MessageDigest.isEqual` (constant-time).
- **Rotation window**: a multi-value `hostKeyFingerprint` (comma/newline separated) accepts either the current or next key — covered by `serverKeyVerifierSupportsHostKeyRotationWindow`.
- **Regression fixed**: a prior head treated the whole multi-line value as a single fingerprint; it's now parsed into a list (`hasValidHostKeyFingerprints` + `SHA256_FINGERPRINT_PATTERN`).
- **Tests**: 15 passing across `FtpProtocolTest`, `FtpCollectImplTest`, `SftpHostKeyCompatibilityMigrationTest`.

### Minor (non-blocking)
- The `insecureSkipVerify=true` path is the explicit dangerous opt-in; ensure the emitted warning is loud enough in logs for operators who enable it. The body says warnings are emitted — good.

No blocking issues. Solid security improvement.

Duansg:
Re-reviewed the current head (`ede8be9`). The description and the code now agree on fail-closed, and every blocking item from my previous two rounds is resolved — verified each one:

- **`depend: ssl` boolean wiring**: `onParamBooleanChanged` now ends with `onDependChanged(String(enabled), field)`, and the existing `ngOnChanges` init dispatch plus `fieldValues.map(String)` make initial visibility, toggle visibility, hidden-value clearing, and the `hasAdvancedParams` panel collapse all work on both the new and edit paths, with spec coverage. It also fixes a pre-existing bug where a persisted `"false"` string was treated as truthy and rewrote port 21→22 on merely opening the edit form.
- **`isInvalid()` dead code**: `preCheck` delegates to `validationError()` — one source of truth.
- **Fingerprint verification**: `KeyUtils.checkFingerPrint(expected, BuiltinDigests.sha256, key)` with the explicit digest; the regex-vs-manual-compare, `limit: 2048`, the global-state-mutating test, and the null-session test are all gone. Insecure-skip warnings are deduplicated per endpoint.
- **Upgrade path**: fail-closed is chosen and documented in `upgrade.md`. CI is green.

That said, one real regression remains, and it breaks exactly the population the docs promise is safe.

## Strict boolean validation meets unreplaced placeholders — legacy plain-FTP monitors go down

`validationError()` has:

```java
if (isNotBlank(ssl) && ssl is neither "true" nor "false")                 → error   // ①
if (isNotBlank(insecureSkipVerify) && neither "true" nor "false")         → error   // ② — before the ssl gate
if (!"true".equalsIgnoreCase(ssl)) return null;                                     // plain FTP returns here
```

But `CollectUtil.replaceSpecialCharacterIfNeeded()` handles a **missing** configmap key with `continue` — the `^_^field^_^` placeholder is left in the value verbatim, not nulled. So any monitor whose param rows lack the new fields delivers the literal `^_^insecureSkipVerify^_^` into check ②.

Which monitors lack rows? I traced the three job-assembly paths:

| Path | Backfills missing params from `defaultValue`? | Consequence |
|---|---|---|
| Startup `SchedulerInit.run()` | Yes (`insecureSkipVerify` has `defaultValue: false`) | Legacy plain FTP survives — **except** monitors created before #3038 (Feb 2025) that were never edited: they have no `ssl` row, `ssl` has **no** defaultValue, the placeholder hits ① and they **fail at startup** |
| `addMonitor` — including `importConfig` (`MonitorServiceImpl` builds configmap without backfill; `validate()` only rejects missing *required* params, it never backfills optional ones) | No | Importing a 1.8.x export of a **plain FTP** monitor → placeholder hits ② → "Sftp Protocol skip-verification option must be true or false." The monitor is down until the next manager restart backfills it — a confusing self-healing failure referencing an SFTP field the operator never saw |
| `modifyMonitor` | No | External API edits without the new fields: same as above |

This contradicts the new `upgrade.md` line "Plain FTP monitors are unchanged." The UI paths are unaffected (the form always submits all param rows, and a null-valued row is correctly replaced to null).

**Suggested fix** (small): drop the strict must-be-true-or-false rejection in ① and ② and use `"true".equalsIgnoreCase(...)` semantics (anything else = false/disabled), or at minimum move ② behind the `ssl == true` gate. Security is unaffected — a placeholder or garbage value never equals `"true"`, so it can never earn the skip-verification exemption; fail-closed is preserved. What's lost is only an error message for explicitly malformed values; what's gained is not breaking existing monitors.

## Minor (non-blocking)

1. A legacy SFTP monitor at startup fails with "must use the SHA256:base64 format" — the placeholder makes `hostKeyFingerprint` non-blank, so the "fingerprint is required" message is never reached, and the operator is told their (never-entered) value has the wrong format. Falls out of the same lenient-parsing fix.
2. The regex allows a trailing `=` (`{43}=?`), but MINA's `getFingerPrint` emits unpadded 43-char base64 — a pasted padded fingerprint passes format validation yet can never match, yielding only the generic "did not match" warning. Drop the `=?` or strip padding before comparison.
3. `INSECURE_WARNING_ENDPOINTS` only grows; past 1024 distinct endpoints new insecure connections log at debug only, and entries are never reclaimed when a monitor is deleted or gains a fingerprint. Edge-scale only.
4. `serializationDoesNotExposeComputedValidationProperties` asserts on `"parsedHostKeyFingerprints"`, but the method is named `parseHostKeyFingerprints` — that assertion can never fire (the `validationError` one is real).

The core of the PR — the verifier, explicit SHA-256, the rotation window, the fail-closed contract, and the form wiring — is solid now. Fix the strict-boolean/placeholder interaction (a few lines) and tune the message, and I'm happy to approve.

Duansg (approved):

PR #4295, opened 2026-07-30
maintenance: scope quick-start listeners

zqr10159:
## What changed?

- merge the latest `master` into this standalone PR;
- bind the quick-start HertzBeat web/API listener (`1157`) and manager/collector transport (`1158`) to `127.0.0.1` by default;
- bind OTLP/gRPC ingestion (`14317`) to loopback by default, with a separate `HERTZBEAT_OTLP_BIND_ADDRESS` override so remote Collector access does not implicitly expose telemetry ingestion;
- keep the PostgreSQL and GreptimeDB host-published development endpoints on loopback;
- document port purposes, remote Collector and OTLP setup, wildcard-bind risk, credential replacement, TLS, and source-network controls in the English and Chinese quick-start guides;
- add `.env.example`, ignore the operator's `.env`, and add focused CI that validates the rendered Compose model for default and explicit override cases.

## Upgrade behavior

Ordinary local quick-start users do not need to change anything. Remote Collectors must set `HERTZBEAT_BIND_ADDRESS` to a manager address they can reach and limit `1158` to Collector source networks. Remote web/API access uses `1157`, preferably behind a TLS reverse proxy. Remote OTLP senders must opt in separately with `HERTZBEAT_OTLP_BIND_ADDRESS` and restrict `14317` to trusted sources. Datastore ports remain loopback-only under both overrides.

This branch targets the latest `master` and has no dependency on another PR.

## Regression proof

The outcome test was first run against the unscoped mappings and failed because the rendered host bindings were not loopback-scoped. While merging current `master`, the same test was extended for the newly published `14317` endpoint and failed on its raw host mapping before the separate OTLP bind variable was added.

## Validation

- `sh script/ci/check-quickstart-compose.sh` — passed for default loopback bindings and independent `HERTZBEAT_BIND_ADDRESS=192.0.2.10` / `HERTZBEAT_OTLP_BIND_ADDRESS=192.0.2.20` overrides; PostgreSQL and GreptimeDB remained loopback-only in both renders;
- wildcard render with both bind variables set to `0.0.0.0` — passed and confirmed all three HertzBeat ports use the explicit override;
- Ruby YAML parsing for the Compose file and workflow — passed;
- `sh -n script/ci/check-quickstart-compose.sh` — passed;
- `git diff apache/master --check` — passed;
- OCR preview and resolved workflow/YAML/default review rules — no remaining high- or medium-severity findings.

Human validation: rendered and inspected the final Compose model for default, independent override, and wildcard-address cases.
Risk notes: remote quick-start access is an explicit operator choice. Operators still need to replace bundled/default credentials and configure network and TLS controls before any non-loopback exposure.


zqr10159:
Author remediation update:

This PR is now standalone and directly targets master; it no longer depends on #4258. All quick-start host-published ports are loopback-scoped by default. HERTZBEAT_BIND_ADDRESS changes only HertzBeat 1157 and 1158, while PostgreSQL and GreptimeDB remain local. English and Chinese docs and .env.example cover the web and API port, manager and collector transport, remote collector setup, wildcard-bind risk, credential replacement, TLS, and source-network controls.

The executable Compose contract first failed because the datastore mappings had no host_ip, then passed for default, documentation-address, and wildcard-address renders. A path-scoped GitHub Actions workflow now checks the final Compose JSON. GitHub checks are still starting on the rebased head; maintainer review remains required.

zqr10159:
CI follow-up: the dedicated quickstart-config job passed, and backend build, Maven E2E, image E2E, license, and label checks have all completed successfully on the rebased current head.

Duansg:
Did a deep review of this PR (diff-first, independent of the description). Overall the change is technically solid — I found no blocking technical defect — but there are two substantive points that I think deserve maintainer consensus before merge, plus a few minor nits. Details below.

## Verified correct

- `${HERTZBEAT_BIND_ADDRESS:-127.0.0.1}` / `${HERTZBEAT_OTLP_BIND_ADDRESS:-127.0.0.1}` interpolation is correct (`:-` also falls back on empty string, which is the right choice over `-`). Compose auto-loads the sibling `.env` for interpolation, so the `.env.example` → `.env` flow works as documented.
- The CI script asserts loopback bindings for `postgres` (5432→15432) and `greptime` (4000–4003→14000–14003) which this diff does not touch — I cross-checked current `master`: those services already carry the `127.0.0.1:` prefix, so the assertions match reality rather than assuming it.
- `check-quickstart-compose.sh`: POSIX sh, `set -eu`, `mktemp` + trap cleanup, `unset` before the default render, and the jq types match `docker compose config --format json` output (`target` numeric via `--argjson`, `published` string via `--arg`). The `quickstart-config` job passing on the real runner confirms this empirically.
- Workflow hygiene is good: `permissions: contents: read`, path filters, concurrency group, 5-minute timeout. `actions/checkout@v4` is consistent with the other 7 usages in this repo.
- All new files carry ASF license headers; license check is green.

## Points that need a maintainer decision

**1. Defaulting the quick-start to loopback is a user-facing behavior change, not pure maintenance.**
After this merges, a user who runs `docker compose up -d` on a remote server can no longer reach `http://<server-ip>:1157` from their workstation browser — which is arguably the most common deployment pattern for a monitoring system. The security motivation is real and well-founded (bundled `admin/hertzbeat` credentials, and Docker-published ports bypassing ufw/firewalld), so I'm sympathetic to the change — but it currently ships with only the in-directory READMEs updated. I'd suggest a release-note / upgrade-doc entry so operators aren't surprised, and an explicit ack from maintainers that the community accepts secure-by-default here.

**2. Consistency gap: only 1 of 5 compose variants is scoped — and not the officially recommended one.**
`home/docs/start/docker-compose-deploy.md` recommends `hertzbeat-postgresql-victoria-metrics`, which still publishes `1157/1158/14317` on all interfaces — and even its `15432:5432` PostgreSQL mapping has no loopback prefix. That same doc also instructs users to "Access http://ip:1157/". So the security benefit doesn't reach the documented default path, and after merge the variants behave differently from each other and from the website docs. Not necessarily a blocker for this PR, but I'd like to see a follow-up issue tracked to align the remaining four variants and the website docs — otherwise this hardening stays an island.

## Minor nits (non-blocking)

- Assertion failures are silent: `jq -e` output goes to `/dev/null`, so when `set -e` aborts, the CI log doesn't say which binding failed. A small error message per assertion would help.
- Running the script locally with a `.env` present in the compose directory pollutes the "default" render (the script `unset`s env vars but can't shield the `.env` file) and produces a false failure. CI is unaffected (clean checkout); `--env-file /dev/null` would make it robust locally too.
- The workflow's path filter covers all of `script/docker-compose/**`, but the script validates only the greptimedb variant — changes to the other variants trigger the job without being checked.
- `host_ip` only accepts an IP literal; a user following the docs' wording "address" might supply a hostname and get a Compose error. Perhaps say "IP address" explicitly in `.env.example`/README.

## Suggested path to merge

Maintainer ack on the loopback-by-default decision → file the follow-up for the remaining variants + website docs + release note → update branch (currently behind master) → approve & merge. If the community prefers keeping the old default, the variable mechanism here is still fully reusable with `0.0.0.0` as the default.

Aias00 (approved):
Approving. This is a clean, well-scoped secure-by-default hardening of the quick-start Docker Compose listeners, with CI that actually proves the bindings.

## Verified
- **HertzBeat listeners default to loopback.** All five variants now map `1157`/`1158` to `${HERTZBEAT_BIND_ADDRESS:-127.0.0.1}` and `14317` (OTLP/gRPC) to the *independent* `${HERTZBEAT_OTLP_BIND_ADDRESS:-127.0.0.1}`. Splitting OTLP onto its own variable is the right call — enabling a remote Collector no longer implicitly exposes telemetry ingestion.
- **Datastore dev ports loopback-only.** PostgreSQL (`15432`), MariaDB/MySQL (`13306`), IoTDB (`18181/16667`), TDengine (`16041`), GreptimeDB (`14000–14003`), VictoriaMetrics (`18428`) are all published on `127.0.0.1`. Containers still reach each other over the internal `hertzbeat` network, so local use is unaffected.
- **No wildcard gap in `postgresql-greptimedb`.** The `0.0.0.0:4000–4003` strings in the `greptime` service are GreptimeDB *intra-container* `command` args (`--http-addr`/`--rpc-bind-addr`/`--mysql-addr`/`--postgres-addr`), not published ports. The host-published ports there are `127.0.0.1:14000–14003`. Confirmed there is no `0.0.0.0` host exposure.
- **CI validation is real and strong.** `script/ci/check-quickstart-compose.sh` renders each variant twice (default + override) and asserts via `jq`: every default binding is `127.0.0.1`; under override, `1157/1158` → `$HERTZBEAT_BIND_ADDRESS`, `14317` → `$HERTZBEAT_OTLP_BIND_ADDRESS`, and all non-hertzbeat (datastore) bindings stay `127.0.0.1`. `set -eu` makes a failed assertion fail the run. The workflow runs it on push/PR and is correctly `permissions: contents: read`.
- **Operators can't leak secrets.** `.env.example` added (or appended) for all five variants with both bind vars; `.gitignore` ignores `.env` in the four variants that lacked it (postgresql-victoria-metrics already had one).
- **Docs are accurate.** `upgrade.md`, `script/docker-compose/README.md`, and the per-variant EN/CN READMEs correctly describe the new default, the two override variables, and the credential/TLS/network controls required before any non-loopback exposure.

## Non-blocking notes
- **Behavior change for existing remote deployments.** Setups that previously reached the manager on `0.0.0.0:1158` will stop working until `HERTZBEAT_BIND_ADDRESS` is set. This is documented, but it's worth a one-line callout in the 1.9.0 release notes so operators aren't surprised.
- **CI environment assumption.** The workflow relies on `docker` (compose plugin) and `jq` being preinstalled on `ubuntu-latest` without an explicit setup step. They are today, but adding `docker/setup-docker-action` would make the dependency explicit and future-proof. Purely a robustness suggestion.
- **Collector-in-same-compose still works** because it reaches `hertzbeat:1158` over the internal network, not `127.0.0.1` — consistent with the README explanation.

No blocking issues. Solid security improvement with regression protection. LGTM.

PR #4302, opened 2026-08-04
Consolidate entity-free observability for 1.9

zqr10159:
## Summary

- replace the log-specific module boundary with `hertzbeat-observability` and move entity-free signal storage/query ownership into the warehouse layer
- expose authenticated OTLP metrics, logs, and traces ingestion over both HTTP and gRPC while keeping the existing observability query routes
- add transition and Greptime end-to-end contracts, update packaging/CI/Sureness configuration, and align the 1.9 integration documentation
- redesign the metrics, logs, and traces pages around a shared dense investigation workspace
- support Greptime-normalized OTLP resource keys in log queries and allow cold PromQL queries to complete without a false storage-unavailable result

## Motivation

The 1.9 release is the transition release before the 2.0 entity model. It needs the mature three-signal ingestion and query boundary without introducing the entity layer. This change consolidates those capabilities under observability and warehouse ownership while preserving the 1.9 operator routes and resource-based filtering model.

## Validation

- HTTP OTLP JSON and gRPC OTLP protobuf smoke against a real local HertzBeat backend and GreptimeDB: each transport independently wrote and queried one metric, one linked log, and one root trace
- browser verification on the production UI routes: both transports displayed the expected metric Series, canonical service-labelled log, root operation, duration, status, and Trace ID; service/environment/operation filtering was exercised
- focused backend tests: 25 warehouse tests, 22 observability tests, one Greptime cold-query timeout contract, and six startup transition contracts
- `./mvnw -pl hertzbeat-startup -am -DskipTests package`: all 24 reactor modules succeeded and produced the 1.9 startup jar
- focused Angular tests: 16/16 passed
- `npm run package`: production build succeeded
- focused ESLint, Stylelint, Prettier, touched locale-key parity across five catalogs, changed-source CJK scans, and `git diff --check` passed

## Upgrade impact

- the Maven module and Java package boundary changes from `hertzbeat-log` to `hertzbeat-observability`
- Greptime pipeline/table initialization is now owned by `hertzbeat-warehouse`
- the 1.9 transition remains resource-scoped and does not require entity identifiers
- existing public observability/OTLP route families remain available; authentication rules are updated consistently in packaged Sureness configurations

Human validation: ran the commands and real HTTP/gRPC/Greptime/browser checks listed above; all required checks passed.
Risk notes: the live local smoke used the cached GreptimeDB 1.0.1 image while repository Compose targets 1.1.3, so CI/E2E remains the final pinned-version proof. The Angular production build retains the repository's existing 3.12 MB initial-bundle budget warning. Existing unrelated full-catalog locale drift was not expanded; all keys added here are present in all five touched catalogs.


Aias00 (dismissed):
## Review: Consolidate entity-free observability for 1.9

**Verdict: ✅ APPROVED (scoped review)** — large, well-tested refactor; high-risk surfaces check out.

> Scope note: this is a +3,149 / −1,488, 123-file change. I did a **scoped** review focused on the risk areas rather than a full line-by-line pass. The author's validation matrix (e2e against real GreptimeDB over HTTP+gRPC, 25 warehouse + 22 observability tests, transition contracts, UI tests, full reactor build) is strong.

### What I verified
- **Module rename is complete**: no leftover `hertzbeat-log` references in non-test sources (only in assertions that prove the rename). Build should not have dangling imports. There's even a `ContextTest` asserting `ThreeSignalQueryController`/`OtlpSignalController` now resolve from `hertzbeat.observability`.
- **Auth config is coherent with the code changes**:
  - `/api/otlp/**===post` narrowed to `/api/otlp/v1/**===post`; old anonymous `/api/observability/capability===get` was removed from `excludedResource` and now requires a logged-in (guest+) role — a tightening, not a regression.
  - Removed GET rules (`/api/ingestion/otlp/metrics/console`, `.../inventory`) correspond to endpoints that were deleted in the same PR — rules and code stay in sync.
  - Deprecated `/api/logs/otlp/**` is kept as an alias forwarded to the canonical `/api/otlp/v1/logs`, with a note it's removed in 2.0.
- **Dependencies**: only `hertzbeat-observability` / `hertzbeat-observability-e2e` are added; no new third-party deps introduced at a glance.
- **GreptimeDB integration** is covered by a real `GenericContainer` e2e test (HTTP + gRPC write/query), and `GREPTIME_QUERY_READ_TIMEOUT` was bumped 5s→15s with a rationale.

### Non-blocking suggestions
1. **Greptime schema files**: the e2e test references `greptime/tables/hertzbeat_logs.sql` and `greptime/pipelines/hertzbeat_otlp_log_v1.yaml`. Please confirm these resource files are added in this PR (the e2e would fail in CI otherwise, so CI is the backstop — just worth a manual glance).
2. **Broad GET rule**: `/api/observability/**===get===[admin,user,guest]` consolidates the former per-signal guest-read rules. Please double-check no sensitive GET endpoint under `/api/observability` is now unexpectedly guest-readable.

No blocking issues found in the scoped review. Good work on the transition boundary.

Aias00 (approved):
## Review: APPROVE

Solid consolidation of the entity-free observability path for 1.9, and a clear net security win on the OTLP ingestion boundary. Re-reviewing because the previous approval was dismissed and the PR grew from +3149/−1488 (123 files) to +4156/−1523 (141 files) with several hardening commits since.

### What this PR does
- **Module consolidation**: `hertzbeat-log` → `hertzbeat-observability`; entity-based multi-protocol log ingestion is replaced by an OTLP-only entity-free path.
- **Canonical routes**: `POST /api/otlp/v1/{logs,metrics,traces}` (ingest, `[admin,user]`), `GET /api/observability/**` (query, `[admin,user,guest]`), `DELETE /api/observability/logs` (admin-only). Deprecated 1.8 aliases (`/api/logs/otlp/**`, `/api/logs/ingest/{protocol}`) are kept as `@Deprecated(forRemoval=true)` forwards that stamp `Deprecation`/`Link` headers, scheduled for 2.0 removal.
- **OTLP/gRPC listener** (default `14317`, off the OTel-standard `4317`) with a bearer-token interceptor, and **graceful degrade** when the port cannot be bound (HTTP ingestion keeps working).

### Security / robustness verified
- **OTLP/HTTP error handler** (`OtlpHttpExceptionHandler`) is correctly scoped (`@RestControllerAdvice(assignableTypes=...)`, `@Order(HIGHEST_PRECEDENCE)`) and **never echoes internal detail**: unexpected exceptions return a generic message; `RestClientException` returns a generic "storage unavailable" without transport/endpoint specifics; the JSON fallback properly escapes the message. It maps to spec-compliant `google.rpc.Status` (JSON or protobuf per request) with correct `429 RESOURCE_EXHAUSTED` / `503 UNAVAILABLE` + `Retry-After`. The old inline handler that echoed `exception.getMessage()` into the body is removed — a real info-leak fix.
- **GreptimeDB write path** (`GreptimeOtlpSignalStorage.writeProtobuf`): distinguishes retryable 4xx (408/425/429, kept retryable) from fatal 4xx (rejected payload → `IllegalArgumentException` → 400 so exporters stop replaying dead batches), bounds the error detail returned to exporters (≤200 chars, full body only logged), and validates the signal name against `{metrics,logs,traces}`. Credential handling is centralized (config-sourced BasicAuth).
- **gRPC** routes logs through the same canonical `logIngestionService.ingestProtobuf` path as HTTP, wrapped in the workload guard, behind the token interceptor.
- **Trace query overflow** fixed: `clampDurationMs` bounds `(0, Long.MAX_VALUE/1_000_000)` so `duration_nano >= x * 1_000_000L` cannot wrap, and `OFFSET` is a `(long)` multiplication.
- **SSE scoping** for alert/manager/log streams is now consistently auth-required (the manager test `sureness.yml` is aligned with production and #4272); `/api/observability/capability` is no longer anonymous.
- Self-telemetry tables renamed `hzb_logs`/`hzb_traces` → `hzb_internal_logs`/`hzb_internal_traces` to avoid colliding with user-facing tables.

### Non-blocking notes
- **Behavior change**: the generic `/api/logs/ingest/{protocol}` controller (vector/loki via `LogProtocolAdapter`) is removed; the legacy alias only accepts `otlp`. Confirm vector/loki ingestion is intentionally dropped and called out in the 1.9 release notes.
- `OtlpAccessTokenValidator` fallback (when no bean is available) rejects every gRPC token. Pre-existing pattern — just ensure the validator bean is always wired in production profiles so gRPC OTLP isn't silently closed.
- GreptimeDB write uses `RestTemplate` BasicAuth over internal HTTP (unchanged from before). Fine on a trusted network; ensure TLS if GreptimeDB is remote.
- Query-side SQL still builds via string concatenation, but relies on `escapeSql`/`SAFE_IDENTIFIER` patterns and `ResourceFilterExpression.parse` validation; the `resourceFilter` key is interpolated into the JSON path. Worth a follow-up fuzz check, but out of scope here (preserved behavior).

Tests are extensive (40 test files, ~+1577 lines) including E2E and an `EntityFreeObservabilityTransitionContractTest` that pins the module rename and the absence of old entity-based routes.

No blocking issues — approving.

Duansg (approved):

PR #4304, opened 2026-08-09
fix(web-app): clear dependent param value when condition not met (#2298)

### LINKED ISSUE #2298: [BUG] When http method change by POST to GET,the request body didn't delete

DSingh0304:
## What's changed?

Fixes #2298. 

**Root Cause:**
In `MonitorFormComponent`, when `onDependChanged` hid dependent fields (such as `payload` when switching `httpMethod` from `POST`/`PUT` to `GET`), it set `display = false` without clearing `paramValue`. As a result, the previous payload data remained in the form state and was submitted on detect/save, causing HTTP 415/400 errors on GET requests.

**Changes:**
1. Updated `onDependChanged` to reset `param.paramValue = null` whenever a dependent field's condition is not met.
2. Dynamically recomputed `hasAdvancedParams` to correctly reflect the advanced fields' visibility state.
3. Added `(ngModelChange)` event binding for `advancedParamDefines` to ensure changes in advanced fields also trigger dependency checks.
4. Added unit tests in `monitor-form.component.spec.ts` covering dependency hiding and clearing behavior.

## Checklist

## Add or update API


Duansg (changes requested):
Hi, please review and resolve the conflict.

DSingh0304:
> Hi, please review and resolve the conflict.

Hi the PR is up to date..

Aias00 (dismissed):
## Review: fix(web-app): clear dependent param value when condition not met (#229)

**Verdict: ✅ APPROVED** — clean UI fix with tests.

### What this PR does
When a dependent param's condition is no longer met, the value is now cleared (`paramValue = null`) instead of only hiding it (`display = false`). This prevents stale values from being submitted for fields that are no longer applicable (e.g., switching HTTP method away from POST/PUT hides the `payload` field and now also wipes its value).

- `onDependChanged`: in all three define lists (`paramDefines`/`sdDefines`/`advancedParamDefines`), the else-branch now sets `display=false` **and** `paramValue=null`.
- Added null-safety (`?.` and `&& this.params && this.params[index]`) guarding against undefined arrays / index mismatch.
- `ngOnChanges` resolves the radio/boolean value from `paramValueMap` → `params` → `defaultValue`, and recomputes `hasAdvancedParams` consistently.
- Two spec tests cover both the clear-when-hidden and show-when-met paths.

### Assessment
- **Correctness**: The fix addresses the actual bug (leftover `paramValue`). The null-safety guards are a good addition and reduce the chance of NPE/undefined-index errors during form init.
- **Tests**: Both new cases assert the new behavior directly. Good.

### Minor (non-blocking) notes
- The code relies on parallel-index coupling between each `*Defines` array and the corresponding `*Params` array (`this.params[index]`). This is a pre-existing assumption, but if the two arrays ever diverge in order/length, the wrong param could be cleared. Worth a follow-up to key params by `field` instead of index.
- Toggling a dependency back and forth will clear-and-restore the value (intended per the issue), which is correct but may surprise users who expected the value to persist. Acceptable.

No blocking issues.

Aias00:
thx for your contribution, pls fix ci

Duansg (approved):

Aias00:
LGTM, thx for your contribution

PR #4308, opened 2026-08-09
[fix] keep one-row metric data when command output is short

### LINKED ISSUE #1560: Use the grep -o command to match fields. If the fields are not matched, the collection system will report an error.

### Current Behavior

采集命令为：df -hT | grep -o "centos-hermitlv"
监控模板为
- field: nfs_mount
type: 1
i18n:
zh-CN: NFS 挂载点
en-US: NFS Mounted

### Expected Behavior

当服务器上卸载centos-hermitlv 监控应该能检测到空字符串，并显示在监控上进行报警

### Steps To Reproduce

Is there an existing issue for this?
 I have searched the existing issues
Current Behavior
输入图片说明
采集命令为：df -hT | grep -o "centos-hermitlv"
监控模板为
- field: nfs_mount
type: 1
i18n:
zh-CN: NFS 挂载点
en-US: NFS Mounted

Expected Behavior
输入图片说明
当服务器上卸载centos-hermitlv 监控应该能检测到空字符串，并显示在监控上进行报警

Steps To Reproduce
1、服务器上创建一个添加到centos的hemitlv lvm卷 并格式化为xfs 挂载到/hermit 目录
2、配置自定义监控模板添加字段

field: nfs_mount
type: 1
i18n:
zh-CN: NFS 挂载点
en-US: NFS Mounted
检测命令为:df -hT | grep -o "centos-hermitlv"
3、当挂载存在时，即能检测到centos-hermitlv，采集正常，当挂载不存在时，即采集直接不显示并且后台有报错参数不足

@

### Environment

```markdown
HertzBeat version(s):1.44
```

### Anything else?

_No response_

orangeCatDeveloper:
## What's changed?

Fixes #1560.

**Scenario**: an SSH/script monitor uses `oneRow` parsing — each output line fills one field. The reporter's script ends with an existence check:

```bash
echo <cpu>; echo <memory>; echo <disk>
df -hT | grep -o "centos-hermitlv"   # mount present -> prints its name; gone -> prints nothing, exits 1
```

When the mount disappears, the script outputs 3 lines instead of 4 — and the user expects the last field to show empty so they can alert on it.

**What actually happened** — two code paths, both losing the data:

1. **Partial output** (multi-field template, the reporter's case): `parseResponseDataByOne` rejected the whole row when output lines were fewer than alias fields, discarding the cpu/memory/disk values that *were* collected. The metric page went blank and every cycle logged `ssh response data not enough`. Ironically, the loop right below that length check already pads missing columns with `NULL_VALUE` — the check just never let short output reach it.
2. **Empty output** (single-field template): any blank stdout was hard-failed with `response data is null`, treating "grep found nothing" (a meaningful answer) the same as "the command blew up".

**Fix**:

- Partial output: remove the length check; missing trailing fields become `NULL_VALUE`, collected values survive.
- Empty output: if there is no stderr and the exit status is <= 1 (the grep-no-match signature), emit a row of `NULL_VALUE` and succeed. Real failures still fail — and the message now carries the captured stderr or exit code instead of the fixed `response data is null`.
- To tell those apart, both collectors now capture stderr and the exit status, which they previously discarded (ssh dumped remote stderr into the collector's own `System.err`; script never read it — it is drained on a daemon thread to avoid the classic stderr-pipe deadlock).
- The two collectors carried identical copy-pasted one-row parsing; it is merged into `OneRowResponseSupport`. The script collector shares the same defect and is fixed together; no other collector has one-row command semantics.
- Bonus: `SshProtocol` gains an optional `charset` (default UTF-8) for non-UTF-8 remote output, matching what `ScriptProtocol` already had.

**Known tradeoff** (pinned by an explicit test): a command that exits 1 silently with no output and no stderr is indistinguishable from a grep no-match and is accepted as empty data. Requiring exit 0 instead would break the reported use case itself, since grep exits 1 on no match.

With `NULL_VALUE` flowing through, the alert calculator sees the field as `null`, so rules like `!exists(nfs_mount)` can fire on "the mount is gone" — exactly the issue's ask.

## Before / after

Verified against a real sshd container, using the reporter's script shape (`echo 52; echo 35.8033; echo 5%; grep -o 'missing-mount' /proc/mounts`) with 5 alias fields (`cpu, memory, disk, nfs_mount, responseTime`):

**Before** — no metric row is stored; the collector logs the same error as the issue's screenshot, every collect cycle:

```
ERROR SshCollectImpl Line:249 - ssh response data not enough: 52
35.8033
5%
```

**After** — the row is stored with the gap visible as null, and no errors are logged:

```json
"origin": "52",        // cpu
"origin": "35.8033",   // memory
"origin": "5%",        // disk
"origin": null,        // nfs_mount  <- alertable via !exists(nfs_mount)
"origin": "14",        // responseTime
```

## Checklist

## Add or update API

### DIFF
--- a/hertzbeat-collector/hertzbeat-collector-basic/src/main/java/org/apache/hertzbeat/collector/collect/common/OneRowResponseSupport.java
+++ b/hertzbeat-collector/hertzbeat-collector-basic/src/main/java/org/apache/hertzbeat/collector/collect/common/OneRowResponseSupport.java
@@ -1,0 +1,113 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hertzbeat.collector.collect.common;
+
+import java.util.Collections;
+import java.util.List;
+import org.apache.hertzbeat.collector.constants.CollectorConstants;
+import org.apache.hertzbeat.common.constants.CommonConstants;
+import org.apache.hertzbeat.common.entity.message.CollectRep;
+import org.springframework.util.StringUtils;
+
+/**
+ * Shared one-row response handling for command-based collectors.
+ */
+public final class OneRowResponseSupport {
+
+    /**
+     * Parse type where each output line maps to one alias field of a single result row.
+     */
+    public static final String PARSE_TYPE_ONE_ROW = "oneRow";
+
+    private OneRowResponseSupport() {
+    }
+
+    /**
+     * Treat blank stdout without an error signal (no stderr, exit status present and &lt;= 1,
+     * grep-style no match) as valid empty one-row data: append a row of null placeholders so the
+     * metric stays visible and alertable.
+     *
+     * @return true if handled as empty success, false if the caller should report a failure
+     */
+    public static boolean tryAppendEmptyOneRow(String parseType, String stdErr, Integer exitStatus,
+                                               List<String> aliasFields, CollectRep.MetricsData.Builder builder,
+                                               Long responseTime) {
+        if (PARSE_TYPE_ONE_ROW.equals(parseType)
+                && !StringUtils.hasText(stdErr)
+                && exitStatus != null && exitStatus <= 1) {
+            appendEmptyValues(aliasFields, builder, responseTime);
+            return true;
+        }
+        return false;
+    }
+
+    /**
+     * Build the failure message for a command that produced no usable stdout: prefer the captured
+     * stderr, then a non-trivial exit status, otherwise the generic null-data message.
+     */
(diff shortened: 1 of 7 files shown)


Aias00 (approved):
Reviewed the full diff (7 files, +526/-61). Well-engineered fix for #1560, and the shared-helper extraction is a nice cleanup. Approving.

## What's good
- **Bug 1 (partial output)**: removing the `lines.length + 1 < aliasFields.size()` guard and routing through `OneRowResponseSupport.appendResponseValues` means collected values survive and only missing *trailing* fields become `NULL_VALUE` — directly fixes the blank-metric-page symptom.
- **Bug 2 (empty output)**: `tryAppendEmptyOneRow` treats no-stderr + exit≤1 (grep no-match) as a valid empty row, and `buildBlankFailureMessage` now surfaces the real stderr or exit code instead of the fixed "response data is null" — much better diagnosability.
- **Stderr is now captured** in both collectors (previously discarded / dumped to `System.err` via `NoCloseOutputStream`). The script collector drains stderr on a **daemon thread into a thread-safe `StringBuffer` with a bounded `join(1000)`**, which correctly avoids the classic stdout/stderr pipe deadlock — good call using `StringBuffer` (not `StringBuilder`) there.
- `SshProtocol` gains an optional `charset` (default UTF-8), matching `ScriptProtocol`.
- **Tests cover the key matrix**: partial output, silent-empty (exit 1), exit-2-still-fails, stderr-priority failure message, plus unit tests for the shared helper.

## Minor suggestions (non-blocking)
- Add one test where the oneRow alias list includes `RESPONSE_TIME`, to lock the column-placement logic (the loop skips consuming a line for `RESPONSE_TIME` and injects `responseTime`) — currently only covered indirectly.
- For SSH, `channel.getExitStatus()` can be `null` when the server doesn't report one; the code safely falls back to FAIL (conservative — good), but a one-line comment would stop a future reader from "fixing" it.
- Note the intentional behavior change: empty oneRow output now yields a `NULL` row instead of `FAIL`. That matches the issue's ask (so `!exists(field)` alerts can fire), just make sure existing oneRow monitors that relied on FAIL-on-empty are aware.

Approving.

PR #4320, opened 2026-08-15
[feature] support Kafka client SCRAM authentication

### LINKED ISSUE #4209: [Feature] Support Kafka SCRAM authentication in Kafka Client monitoring
## Feature Request

Add SCRAM authentication support to the Kafka Client monitoring template and its Java-side AdminClient configuration.

## Is your feature request related to a problem? Please describe

The current Kafka Client template cannot monitor Kafka clusters that require SASL/SCRAM authentication.

Updating `app-kafka_client.yml` alone is insufficient. The YAML template can add authentication fields to the page and pass their values into the kclient protocol configuration, but the Java code that creates Kafka `AdminClient` must also put those values into Kafka `Properties`.

At present, the official implementation only passes the host, port, and `monitorInternalTopic`. The official documentation does not provide SASL parameters, so the built-in template cannot connect to SCRAM-protected Kafka clusters.

Documentation: https://hertzbeat.apache.org/zh-cn/docs/help/kafka_client/

## Describe the solution you would like

Please support SCRAM authentication end to end:

- Add optional authentication fields to `app-kafka_client.yml`, including username, password, SASL mechanism, and security protocol.
- Pass these values through the kclient protocol configuration.
- When creating Kafka `AdminClient`, populate the corresponding Kafka properties, such as `security.protocol`, `sasl.mechanism`, and `sasl.jaas.config`.
- Support at least `SCRAM-SHA-256` and `SCRAM-SHA-512`.
- Keep the existing unauthenticated behavior backward compatible when authentication is not configured.
- Avoid exposing the password in logs or diagnostic output.
- Update the Kafka Client documentation with a SCRAM configuration example.

## Describe alternatives you have considered

Maintaining a locally modified YAML template and Java implementation is possible, but it diverges from the official template and must be maintained across upgrades. YAML-only customization cannot solve the issue because the final Kafka `Properties` are created in Java.

## Additional context

A typical Kafka client configuration requires values equivalent to:

```properties
security.protocol=SASL_PLAINTEXT
sasl.mechanism=SCRAM-SHA-256
sasl.jaas.config=org.apache.kafka.common.security.scram.ScramLoginModule required username="..." password="...";
```

`SASL_SSL` should also be usable when TLS is required.

goutamadwant:
## What's changed?

Adds optional SCRAM authentication to Kafka Client monitoring.

- add `PLAINTEXT`, `SASL_PLAINTEXT`, and `SASL_SSL` security protocol options
- support `SCRAM-SHA-256` and `SCRAM-SHA-512` with username and password credentials
- configure the Kafka AdminClient with the selected security settings while preserving unauthenticated monitoring by default
- isolate cached AdminClient connections by authentication settings and keep passwords out of protocol diagnostics
- document the new settings in the English and Chinese Kafka Client monitoring guides

Closes #4209

## Verification

- `KafkaProtocolTest` (23 tests)
- `KafkaCollectTest` (7 tests)
- `YamlCheckScript` through the 22-module manager reactor
- repository Markdown lint (567 files)

## Checklist

## Add or update API

Not applicable: this change does not add or update an API. SCRAM validation, AdminClient property mapping, credential escaping, cache isolation, and plaintext compatibility are covered by unit tests.


hertzbeat-common-core/src/main/java/org/apache/hertzbeat/common/entity/job/protocol/KafkaProtocol.java
```diff
@@ -74,6 +113,16 @@ public boolean isInvalid() {
...
         }
-        return false;
+        if (StringUtils.isNotBlank(securityProtocol)
+                && !SECURITY_PROTOCOLS.contains(securityProtocol.toUpperCase(Locale.ROOT))) {
+            return true;
+        }
+        if (!hasSaslAuthentication()) {
+            return StringUtils.isNotBlank(saslMechanism)
+                    || StringUtils.isNotBlank(username)
+                    || StringUtils.isNotBlank(password);
```
Duansg, line 123: Could this cause the existing monitoring to become unavailable after the upgrade?

Duansg (changes requested):

Aias00 (approved):
## Review: [feature] support Kafka client SCRAM authentication

**Verdict: ✅ APPROVED** — well-structured feature with solid test coverage.

### What this PR does
Adds SCRAM (SCRAM-SHA-256/512) authentication support for the Kafka client collector:
- `KafkaProtocol`: new `securityProtocol`, `saslMechanism`, `username`, `password` fields; `password` is `@ToString.Exclude` (no credential leak in logs); `isInvalid()` now validates the security protocol against a whitelist and requires credentials when SASL is selected.
- `KafkaCollectImpl`: builds SASL JAAS config, correctly escapes `\` and `"` via `escapeJaasValue`, and folds credentials into the `CacheIdentifier` so different creds get distinct cached `AdminClient`s.
- `app-kafka_client.yml`: conditional fields (`depend` + `hide`) for SASL, wired into all 4 metric sets.
- Docs (EN/ZH) updated; tests added for both collector and protocol.

### Assessment
- **Security**: Credential escaping + `@ToString.Exclude` are the right calls. The cache key now distinguishes by username/password/security protocol, preventing credential cross-contamination between monitors.
- **Correctness**: `hasSaslAuthentication()` safely handles a null `securityProtocol` (no NPE via `String.equalsIgnoreCase(null)` → false). `isInvalid()` correctly treats a blank protocol as PLAINTEXT (backward compatible) and enforces all three SASL fields when SASL is on.
- **Tests**: Good matrix — no-auth, SCRAM auth, JAAS escaping, cache isolation, whitelist validation, and the `toString`-no-password guarantee.

### Minor (non-blocking) suggestions
- `escapeJaasValue` does not escape `;`, which would prematurely terminate the JAAS statement if present in a credential. Extreme edge case, but worth a comment or an additional escape.
- `password` YAML field has no `limit` (username has `limit: 50`); long passwords are plausible — consider adding one.
- `SASL_SSL` relies on the JVM trust store with no `ssl.*` properties supplied; the doc correctly warns about this, but a follow-up could surface truststore config. Acceptable for this scope.

No blocking issues. Nice contribution.

hertzbeat-common-core/src/main/java/org/apache/hertzbeat/common/entity/job/protocol/KafkaProtocol.java
goutamadwant: @Duansg existing monitors remain backward compatible. Persisted monitor definitions have no values for the new fields, so `securityProtocol` and the SASL fields deserialize blank; `KafkaProtocol.isInvalid()` continues to accept that state, and `getAdminClientProperties()` emits only `bootstrap.servers`, matching the previous behavior.

The YAML default applies `PLAINTEXT` only when creating or editing a definition.

PR #4330, opened 2026-08-17
[fix] allow long threshold expr when binding many monitors

### LINKED ISSUE #4171: [BUG] <title>Association monitoring: Cannot add more than 30 monitoring items (Size.alertDefine.expr 0-2048)

### Current Behavior

Association monitoring: Cannot add more than 30 monitoring items (Size.alertDefine.expr 0-2048)
Actual measurement, correlation monitoring: The number of items that can be added cannot exceed 30, which is far less than the specified maximum of 2048

Image
Image

### Expected Behavior

Adjust to 0-2048

### Steps To Reproduce

_No response_

### Environment

```markdown
HertzBeat version(s):1.8.0
```

### Debug logs

_No response_

### Anything else?

_No response_

orangeCatDeveloper:
Fixes #4171

Binding more than ~30 monitors to one alert rule fails with a validation error, because each bound monitor appends an `equals(__instance__, "...")` clause (~65 chars) to the threshold `expr`, and the field was capped at 2048 characters — 2048 / 65 ≈ 30 bindings.

The fix widens `AlertDefine.expr` from `@Size(2048)/@Column(2048)` to `@Lob` with a `@Size(65535)` validation cap — the same pattern the codebase already uses for `NoticeTemplate.content`. Migration `V182` alters the column on all three vendors (mysql `longtext` with an INFORMATION_SCHEMA idempotency guard matching V172/V180, postgresql `TEXT`, h2 `CLOB`); widening is lossless for existing rows. The custom validation message was deliberately dropped so the default hibernate-validator message stays localized.

Backward compatible: previously valid input remains valid, the new 65535 cap still rejects runaway payloads, and `expr` is admin-authored config so no new trust boundary.

Verified end-to-end against a real backend (fresh H2, `Successfully applied 7 migrations ... now at version v182`):

```
before (2048 cap):
  POST expr=1501 chars (35 bindings)  -> {"code":0,"msg":"Add success"}
  POST expr=2296 chars (50 bindings)  -> {"code":1,"msg":"Size.alertDefine.expr:size must be between 0 and 2048"}
                                          (verbatim the error reported in the issue)

after (lob + 65535 cap):
  POST the same 2296-char expr        -> {"code":0,"msg":"Add success"}
  read back                           -> expr length 2296, stored intact
  POST 65536 chars                    -> {"code":1,"msg":"Size.alertDefine.expr:size must be between 0 and 65535"}
```

### DIFF
--- a/hertzbeat-common-spring/src/main/java/org/apache/hertzbeat/common/entity/alerter/AlertDefine.java
+++ b/hertzbeat-common-spring/src/main/java/org/apache/hertzbeat/common/entity/alerter/AlertDefine.java
@@ -69,8 +69,8 @@ public class AlertDefine {
     private String type;
 
     @Schema(title = "Alarm Threshold Expr", example = "usage>90", accessMode = READ_WRITE)
-    @Size(max = 2048)
-    @Column(length = 2048)
+    @Size(max = 65535)
+    @Column(columnDefinition = "TEXT")
     private String expr;
 
     @Schema(title = "Execution Period/ Window Size (seconds) - For periodic rules/ For log realtime", example = "300")
--- a/hertzbeat-common-spring/src/test/java/org/apache/hertzbeat/common/entity/alerter/AlertDefineTest.java
+++ b/hertzbeat-common-spring/src/test/java/org/apache/hertzbeat/common/entity/alerter/AlertDefineTest.java
@@ -1,0 +1,72 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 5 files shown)


Aias00 (dismissed):
## Review: allow long threshold expr when binding many monitors

Clean, well-migrated fix for #4171 (binding >~30 monitors to one rule failed validation).

**What's good**
- `AlertDefine.expr` is widened from `@Size(max=2048) @Column(length=2048)` to `@Lob` with a `@Size(max=65535)` validation cap — the same pattern the codebase already uses for `NoticeTemplate.content`. Previously valid input stays valid; the new 65535 cap still rejects runaway payloads. `expr` is admin-authored config, so this widens no new trust boundary.
- Migrations `V182` for all three vendors are correct and idempotent:
  - **MySQL**: a stored procedure guarded by `INFORMATION_SCHEMA` (table exists, column exists, and `DATA_TYPE != 'longtext'`) then `ALTER ... MODIFY COLUMN expr LONGTEXT` — matching the existing `V172`/`V180` style, so re-running is a no-op.
  - **PostgreSQL**: `ALTER COLUMN expr TYPE TEXT` — lossless widen.
  - **H2**: `ALTER COLUMN expr CLOB`.
- Tests pin both directions: a 76-monitor-binding expr (>2048 chars) validates cleanly, and a 65536-char expr fails with exactly one violation on `expr` (the custom message was intentionally dropped so the default localized hibernate-validator message is used).
- Backward compatible: existing rows are widened losslessly, and the PR's e2e trace confirms a 2296-char expr now persists intact.

**Non-blocking note**
- MySQL `longtext` can hold far more than 65535 chars, but the `@Size(65535)` validation is the effective ceiling, so there's no practical over-allocation concern.

No blocking issues. Approve.

orangeCatDeveloper:
Rebased to solve a conflict @Aias00

Aias00 (approved):

PR #4331, opened 2026-08-17
[fix] never send null template variables to Aliyun SMS

### LINKED ISSUE #4319: [Question] <title>默认部署后，阿里云短信发送失败
### Question

默认部署后的几个问题：
1. 短信模板是否只能用一个，不能更换，不能每个项目用独立的模板吗？
2. 为什么在配置完阿里云的ak及短信模板code等信息后，通知媒介添加接收对象时候，测试总是失败，报错“模板变量JSON格式错误 或 JSON变量属性与模板占位符不一致”
3. 文档中关于短信通知配置方式很难看懂，并且给出的短信模板压根不对，阿里云的占位符应该是${xxx}，但是文档给的是    监控项：`instance`，告警级别：`priority`。内容：`content`
这里我按照阿里云的格式配置完还是不能用。
我怀疑不是占位符的问题，应该是content或者哪个属性没有值造成的异常，这种测试发送应该可以手动输入三个占位符的值才对。

orangeCatDeveloper:
Fixes #4319

Sending a test SMS through Alibaba Cloud always fails with Aliyun's "模板变量JSON格式错误 或 JSON变量属性与模板占位符不一致" error, regardless of configuration — and the docs make it worse by showing a template example in the wrong placeholder syntax.

Root cause: the hardcoded test alert built by `sendTestMsg` carries no `instance` label and no `groupKey`, while `AlibabaSmsClientImpl` assumes real-alert shape and falls through to `templateParam.put("instance", null)`. Aliyun rejects any request whose template-variable JSON contains a null/blank value, so the test path was broken for every user.

Three layers:
- `AlibabaSmsClientImpl`: param building extracted into `buildTemplateParam`, every variable falls back to non-blank text (`instance` → groupKey → "unknown"; `content` → summary → description → any annotation → "alert triggered"). Null can no longer reach the JSON, for test and real alerts alike; also removes an NPE when `commonAnnotations` is null.
- `sendTestMsg`: the shared test alert now carries an `instance` label — all five SMS providers read it.
- Docs (en/zh): the Aliyun template example used backticks; corrected to Aliyun's `${...}` syntax with a note that the template must declare exactly these three variables.

Boundary: per-rule SMS templates (question 1 in the issue) is a feature request, not covered here — suggested as a separate issue in the discussion. No Aliyun account was available for a live send (custom templates require review), so evidence stops at the outbound JSON — which is what Aliyun validates.

Before / after (the exact `TemplateParam` sent to Aliyun for the test alert):

```
before: {"instance":null,"priority":"unknown","content":"Please check the CPU usage of the server"}
        -> rejected: template variable JSON error / placeholder mismatch (the issue error)

after:  {"instance":"127.0.0.1","priority":"unknown","content":"Please check the CPU usage of the server"}
```

Verified: 3 new unit tests (no-null for test-alert shape, no-null for empty alert, real values pass through untouched); full alerter suite 468/468; end-to-end smoke of the shared `send-test-msg` path via the email channel against a local SMTP server (message rendered and delivered).


Aias00 (approved):
## Review: never send null template variables to Aliyun SMS

Clean, well-tested bug fix for the always-failing Aliyun SMS test path (#4319).

**What's good**
- `AlibabaSmsClientImpl.buildTemplateParam` now uses a shared `firstNonBlank(...)` fallback so no template variable can ever be null or blank: `instance` → labels.instance → groupKey → `"unknown"`; `priority` → labels.priority → `"unknown"`; `content` → summary → description → first annotation → `"alert triggered"`. This directly fixes the `{"instance":null,...}` rejection Aliyun returns.
- It also closes a latent NPE: `alert.getCommonLabels()` / `getCommonAnnotations()` are now null-guarded to `Map.of()`, whereas the old code dereferenced `getCommonAnnotations().get(...)` directly.
- `sendTestMsg` now seeds the shared test alert with an `instance` label (`127.0.0.1`) so the test path produces a non-null `instance` for all five SMS providers, not just Aliyun.
- Docs (en + zh-CN) corrected: the Aliyun template example used backticks and is now the correct `${instance}` / `${priority}` / `${content}` syntax, with a note that exactly these three variables must be declared.
- Tests are solid: 3 new `AlibabaSmsClientImplTest` cases (no-null for the test-alert shape, no-null for an empty alert, real values pass through untouched) on top of the full alerter suite (468/468), plus an end-to-end SMTP smoke of the shared `send-test-msg` path.

**Non-blocking note**
- Per-rule SMS templates (question 1 in the issue) is explicitly out of scope here and suggested as a follow-up — reasonable scoping.

No blocking issues. Approve.

PR #2009, opened 2024-05-19
[feature] Support time calculation expressions.

LiuTianyou:
## What's changed?

1. add time calculation expression utils and unit test.
2. update openai monitoring template add time params.
3. the http protocol collection type supports query parameter usage time calculation expressions.

## Checklist

- [x]  I have read the [Contributing Guide](https://hertzbeat.apache.org/docs/community/code_style_and_quality_guide)
- [ ]  I have written the necessary doc or comment.
- [x]  I have added the necessary unit tests and all cases have passed.

## Add or update API

- [ ] I have added the necessary [e2e tests](https://github.com/apache/hertzbeat/tree/master/e2e) and all cases have passed.

I will add relevant documentation in the next PR

Issue #1525, opened 2024-02-01
[Question] <title>自定义模版，无法动态的传递参数


yuanmingcong:
### Question

自定义模版，无法动态的传递参数，例如http请求，上送的payload，无法动态的获取当前时间和一些函数，建议在参数模块和metrics模块新增calculates，进行Google Aviator语法操作

hertzbeat:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
### Question

Custom templates cannot dynamically pass parameters, such as http requests, submitted payloads, and cannot dynamically obtain the current time and some functions. It is recommended to add calculates in the parameter module and metrics module to perform Google Aviator syntax operations @@=== =[Question] <title>Customized template, parameters cannot be passed dynamically

PR #4334, opened 2026-08-18
feat(collector): resolve time expressions in http url and headers

orangeCatDeveloper:
Part of #1525

Time expressions like `${@date-1d}` work in an HTTP monitor's query params and payload, but not in the request path or header values - there they are sent to the target verbatim.

## What's changed?

Run the request path (URL) and header values through the same `TimeExpressionUtil.calculate()` that query params and payload already use. Strings without a `${@...}` placeholder are returned unchanged, so ordinary URLs and headers are unaffected.

Example - an HTTP monitor configured with `URL: /api/report/${@year}` and `Header: X-Date: ${@date}`, collected on 2026-08-18:

```
before  GET /api/report/${@year}     X-Date: ${@date}

after   GET /api/report/2026         X-Date: 2026-08-18
```

Scope note: this only extends the existing time-expression support (added in #2009). The original issue also asks for a general expression engine (arithmetic, functions) - that involves sandboxing user-supplied expressions and is out of scope here.

## Checklist

## Add or update API


Aias00 (approved):
## Review: resolve time expressions in http url and headers

Small, well-scoped extension of existing functionality.

**What's good**
- The request URL (`TimeExpressionUtil.calculate(httpProtocol.getUrl())`) and each header value (`TimeExpressionUtil.calculate(header.getValue())`) now go through the same resolver that query params and body already use. `TimeExpressionUtil.calculate` is idempotent for strings without a `${@...}` placeholder, so ordinary URLs/headers are completely unaffected — no behavior change for existing monitors.
- The URL is resolved once into a local `url` and that value is then used for both the query-string split and URI encoding, so there is no double resolution or inconsistency.
- Good integration test (`collectResolvesTimeExpressionsInUrlAndHeaders`) spins up a real `HttpServer`, asserts the resolved year appears in both the request URI and the `X-Year` header. Docs updated in both English and zh-CN.

**Non-blocking note**
- A time expression that resolves to a URL-special character (e.g. a `?` or `/`) could, in theory, affect the `url.contains("?")` query-string split. This is a pre-existing limitation of the query-param path too and is out of scope for this PR (the issue's general expression-engine part is explicitly excluded). Not a regression.

No blocking issues. Approve.

Issue #4136, opened 2026-05-06
[Question] CPU and Memory Utilization too High for Manager/startup module [Distributed Architecture startup , collector , warehouse]


kaushaldeokar:
### Question

1. Currently Hertzbeat uses apache arrow and hertzbeat has both in-memory heap and external memory usage .
 
It uses apache arrow to write in external memory .
 
**Our scenario :** we have distributed actictecture where we spin multi-collectors and warehouses . data get converted to arrow object in collector after collection and is passed to manager via netty here in manager data gets deserialized and stored into memory and suspect manager memory is increased due to this and later the data send to warehouse for persistence via kafka topic .

if we add more monitors the memory grows drastically -> need to solve this 
 
here manager memory is increased and not decreased later .
 
tried to clear the objects from memory before sending to kafka topic but it didnot work .
 
here is sample code from manager where it pushes the data to kafka .

attached the ss for Heap and Code as well

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/477ca1bb-f44c-4076-939c-9e4a47d79e29" />
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/229fca54-89cd-4c17-afda-6f5e30cef4b6" />
<img width="1708" height="595" alt="Image" src="https://github.com/user-attachments/assets/7bcb2487-7417-451f-b959-f3fafe41057a" />
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/c0816e24-3ec9-4859-a266-1c7e189eab9c" />
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/73a25e66-87b2-4548-9cbb-f4c5e695d4de" />
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/9d1e9908-ec90-44c0-8cc9-238dcaafabf3" />

<img width="790" height="478" alt="Image" src="https://github.com/user-attachments/assets/7b2e6c3f-dd35-4a16-bdb9-df2ed506d772" />

paultanay:
I traced the data path in the source and found what I believe is the root cause.

In ArrowUtil.deserializeMultipleRoots(), every time the Manager receives a Netty message from a collector, a new RootAllocator() and one `ArrowStreamReader` per metrics group are created - neither is ever closed:

```java
RootAllocator allocator = new RootAllocator(); // never closed
[...]

PR #4335, opened 2026-08-19
fix(warehouse): close MetricsData after processing to release Arrow buffers

paultanay:
**What's changed?**

`DataStorageDispatch.startPersistentDataStorage()` polls `MetricsData` objects from the queue and passes them through the history, plugin, and realtime writers, but never called `close()` afterward. `MetricsData` implements `AutoCloseable` and wraps an `ArrowTable` backed by Arrow column vectors - without `close()`, those vectors and their buffers are retained until GC decides to collect them. Under production load with many monitors this compounds into the steady heap growth reported in #4136.

Wrapped the polled object in try-with-resources so `MetricsData.close()` is guaranteed on all exit paths: normal completion, writer exception, and thread interrupt. `DataStorageDispatch` is the terminal consumer, so closing here is safe.

**Fixes:** #4136

---

**Checklist**
- [x] I have read the [contribution guidelines](https://hertzbeat.apache.org/docs/others/contributing)
- [x] Existing tests pass (`hertzbeat-warehouse`, `hertzbeat-common-core`)


Aias00 (approved):
## Review: close MetricsData after processing to release Arrow buffers

Clean, minimal, and correct resource-leak fix.

**What's good**
- `commonDataQueue.pollMetricsDataToStorage()` is now wrapped in try-with-resources, so `MetricsData.close()` (releasing the underlying Arrow column-vector buffers) is guaranteed on every exit path: normal completion, writer exception, and thread interrupt. This directly addresses the steady heap growth reported in #4136.
- `DataStorageDispatch.startPersistentDataStorage()` is the terminal consumer of these `MetricsData` objects, so closing them here is safe — no downstream reader needs the Arrow buffers afterward.
- The original control flow is preserved exactly: the inner `try/finally` still ensures `realTimeDataWriter.saveData(...)` runs even if `calculateMonitorStatus`/history/plugin steps throw. The null-guarded `continue` is still valid under try-with-resources (a null resource is simply not closed).
- `MetricsData implements AutoCloseable`, so the compiler-enforced contract holds.

**Non-blocking note**
- Standard try-with-resources behavior means if `close()` itself throws, it would suppress an earlier exception from the body. That is an acceptable, rare edge for this terminal-sink path.

No blocking issues. Approve.

paultanay:
CI failures are in the setup-deps infrastructure step (fails before any compilation). Build and all 88 tests pass locally under the release profile (mvn clean package -Prelease). Looks like a runner issue.

PR #4338, opened 2026-08-21
[feature][alerter] Support notification template preview

### LINKED ISSUE #1343: [Task] Support notification template preview
### Description

image

Add a "Preview" button to preview confirmed template realtime which will be sent to end user actually.

### Task List

_No response_

Prabal864:
Fixes #1343.

Adds a "Preview" capability for notice templates: render a template's content against a representative sample alert and see the output, without saving the template or sending anything.

### What's new

- `NoticeTemplateRenderer` — extracted the FreeMarker rendering logic that lived inside `AbstractAlertNotifyHandlerImpl.renderContent()` into a shared utility. Both the real notify dispatch path (all `AlertNotifyHandler` implementations) and the new preview endpoint now render a template through the exact same code, so preview output matches what actually gets sent.
- `POST /api/notice/template/preview` — accepts a (possibly unsaved, no `id` needed) `NoticeTemplate`, renders it against a sample alert, returns the rendered text. Errors (e.g. malformed template syntax) come back as a normal `Message.fail` response rather than a 500.

### Why not reuse `/receiver/send-test-msg`?

That endpoint tests a *receiver channel* by actually sending a message through it. This is testing a *template's content*, with no receiver involved and nothing sent — a different concern.

### Testing

- Added `previewNoticeTemplate` / `previewNoticeTemplateWithInvalidContent` to `NoticeConfigControllerTest`, covering the success path and the malformed-template-syntax path.
- Ran the full `hertzbeat-alerter` test suite (459 tests) after the extraction — it caught a real regression on my first pass: pulling `renderContent` into a static utility that re-fetched `ResourceBundleUtil.getBundle("alerter")` internally silently dropped the per-instance `bundle` field that `AbstractAlertNotifyHandlerImpl` subclasses (and their tests) inject/mock. Fixed by passing the `ResourceBundle` through as a parameter instead of having the utility fetch it itself. All 459 tests pass after the fix.
- `checkstyle:check` — 0 violations.


Aias00 (approved):
## Review: Support notification template preview

Solid refactor with a clean new capability.

**What's good**
- The FreeMarker rendering logic is correctly extracted into a shared `NoticeTemplateRenderer.renderContent()` that both the real dispatch path (`AbstractAlertNotifyHandlerImpl`) and the new `/api/notice/template/preview` endpoint now call — so preview output is guaranteed to match what actually gets sent. Good design.
- The renderer keeps `cfg.setNewBuiltinClassResolver(TemplateClassResolver.SAFER_RESOLVER)`, so there is no new template-injection surface (no arbitrary class/object resolution, fixed model). This is consistent with the existing alert-render path.
- The preview endpoint is non-destructive: it renders a (possibly unsaved) template against a fixed sample `GroupAlert` and returns the text; nothing is persisted and nothing is sent. Malformed template syntax returns `Message.fail` (FAIL_CODE) instead of a 500 — correct error handling.
- The regression you caught on the first pass (the per-instance `bundle` field being silently dropped when `renderContent` became a static utility) is properly fixed by passing the `ResourceBundle` as a parameter. 459-module tests + checkstyle pass.

**Non-blocking notes**
1. Worth a quick confirm that `/api/notice/template/preview` is covered by the existing `/api/notice/**` auth scope in `sureness.yml` (the PR doesn't modify it, so it should inherit the existing rule like `send-test-msg` does). Since it renders arbitrary user-supplied template content, it should require the same auth as the rest of `/api/notice/**`.
2. The carried-over `TODO` about `Configuration`/`StringTemplateLoader` reuse per call is a minor perf item, not a correctness issue.

No blocking issues. Approve.

PR #4342, opened 2026-08-22
[fix] prevent duplicate dispatch when collector task times out

### LINKED ISSUE #4203: [BUG] Collector timeout fallback at 4 minutes causes duplicate dispatch when actual collection takes longer

### Current Behavior

Now the collector will be assigned a timeout if it runs over 4 minutes, causing the same job to be executed twice when the actual collection exceeds 4 minutes.

### Expected Behavior

The logic in dispatchCollectData should be executed only once, whether triggered by normal collection completion or timeout fallback — whichever comes first.

### Steps To Reproduce

_No response_

### Environment

```markdown
HertzBeat version(s):
```

### Debug logs

_No response_

### Anything else?

i will fix it

paultanay:
Fixes #4203

**Root cause**

`monitorCollectTaskTimeout()` pre-removed the map entry before calling
`dispatchCollectData()`. If a naturally finishing `MetricsCollect` arrived
in that window, it called `dispatchCollectData()` as well, producing a
duplicate `cyclicJob()` and `sendMetricsData()`.

The Prometheus path in `MetricsCollect.run()` also had no `fastFailed()`
guard, unlike the standard path.

**Fix**

`dispatchCollectData()` now owns the atomic `metricsTimeoutMonitorMap.remove()`
and acts as the once-wins gate. A duplicate call sees `metricsTime == null`
and returns immediately.

For priority > 0 timed-out metrics the monitor removes the entry and cancels
the in-flight collect directly, since no TIMEOUT payload is sent downstream.

`cancel()` fires after `dispatchCollectData()` so the `!isCancelled()` guard
does not block `cyclicJob()`.

Added `fastFailed()` check before the Prometheus list-dispatch in `MetricsCollect`.

**Tests**

New `CommonDispatcherTest`:
- Priority-0 expired entry: dispatches once, cancels, map empty.
- Second scan of same entry: no duplicate dispatch.
- Priority-1 expired entry: cancels without dispatching.
- Non-expired entry: nothing touched.


Aias00 (approved):
## Review: prevent duplicate dispatch when collector task times out (draft)

Solid fix for the duplicate-dispatch race described in #4203. (Marking as draft — reviewing the code, not the WIP state.)

**What's good**
- The fix moves the atomic `metricsTimeoutMonitorMap.remove()` out of the timeout scan loop and into `dispatchCollectData()` itself, which now acts as the once-wins gate: a naturally-finishing `MetricsCollect.run()` that calls `dispatchCollectData()` after the timeout already fired sees `metricsTime == null` and returns early — no second `cyclicJob()`/`sendMetricsData()`.
- The priority split is correct: priority 0 timed-out metrics still dispatch the TIMEOUT payload then cancel (so `cyclicJob()` inside `dispatchCollectData` fires before `cancel()`); priority > 0 timed-out metrics remove the entry and cancel the in-flight collect directly, since no TIMEOUT payload is sent downstream.
- `MetricsCollect.run()` now has the `fastFailed()` guard on the Prometheus path, matching the standard path — closes the second duplicate vector the root cause described.
- `recordCollectMetrics` now reports a distinct `"timeout"` status instead of folding it into `"fail"`, which is better observability.
- `CommonDispatcherTest` is comprehensive: priority-0 dispatches once + cancels + map empty, second scan does not double-dispatch, priority-1 cancels without dispatching, non-expired entry is left untouched.

**Non-blocking note**
- The once-wins early-return guard `if (metricsTime == null && !metrics.isHasSubTask() && metrics.getPrometheus() == null) return;` deliberately excludes subtask/prometheus metrics from the short-circuit. The reasoning is sound (those paths manage the entry differently), but a one-line comment explaining *why* subtask/prometheus results must still be processed when the entry is already gone would help the next reader. Tests don't cover that branch specifically — worth a maintainer glance.

No blocking issues for the logic. Approve (subject to the draft being ready to merge).

PR #4350, opened 2026-08-24
[alerter]bugfix: fix Gotify default webhook URL

### LINKED ISSUE #4349: [BUG] Gotify fallback webhook URL is missing token placeholder

### Describe the bug

`AlerterProperties.java` defines the fallback default Gotify webhook URL as:

```java
private String gotifyWebhookUrl = "https://push.example.de/message?token=";
```

But `GotifyAlertNotifyHandlerImpl` formats this value with the receiver token:

```java
String webHookUrl = String.format(alerterProperties.getGotifyWebhookUrl(), receiver.getGotifyToken());
```

Because the fallback default does not contain a `%s` placeholder, `String.format()` ignores `receiver.getGotifyToken()`. If this Java fallback value is used, the generated URL becomes:

```text
https://push.example.de/message?token=
```

instead of including the Gotify token.

The configured value in `application.yml` already uses the expected format:

```yaml
gotify-webhook-url: http://127.0.0.1/message?token=%s
```

So the Java fallback default should be consistent with the configured default.

### Expected Behavior

The fallback default should include the token placeholder:

```java
private String gotifyWebhookUrl = "https://push.example.de/message?token=%s";
```

### Steps To Reproduce

_No response_

### Environment

```markdown
HertzBeat master branch / 2.0-SNAPSHOT.
```

### Debug logs

_No response_

### Anything else?

_No response_

paultanay:
## What's changed

Fixes #4349

This PR fixes the fallback default Gotify webhook URL in `AlerterProperties`.

`GotifyAlertNotifyHandlerImpl` builds the webhook URL with `String.format(...)`, but the Java fallback default was missing the `%s` placeholder. When that fallback value is used, the receiver Gotify token is ignored and the generated request URL does not include the token.

This change makes the Java fallback default consistent with the configured value in `application.yml`.

## Changes

- Added the missing `%s` token placeholder to the default Gotify webhook URL.
- Added/updated unit test coverage for Gotify URL formatting.

## Test

```bash
mvn -pl hertzbeat-alerter -Dtest=GotifyAlertNotifyHandlerImplTest test
```


paultanay:
ready for review @Aias00 could you please run the ci once, thanks...

PR #4257, opened 2026-07-28
[security] Restrict plugin upload endpoint to admin role (#4149)

Bhavya-Sonigra:
Fixes #4149

## What's changed?

The `/api/plugin` endpoint had no role-based access control in `sureness.yml`, unlike every other API path in the file. Since plugin upload loads custom JAR code into the JVM via ServiceLoader, this meant any authenticated (or in some cases unauthenticated) user could manage plugins with the same access as an admin.

This PR adds explicit admin-only role mappings for `/api/plugin/**`, following the exact pattern used elsewhere in the file:

- /api/plugin/**===get===[admin]
- /api/plugin/**===post===[admin]
- /api/plugin/**===put===[admin]
- /api/plugin/**===delete===[admin]

Tested locally:
- Request to `/api/plugin` with no auth token → `401 Unauthorized`
- Request to `/api/plugin` with an admin token → `200 OK` with expected data

## Checklist
- [x] I have read the Contributing Guide
- [ ] I have written the necessary doc or comment.
- [ ] I have added the necessary unit tests and all cases have passed.

PR #4359, opened 2026-09-02
[doc] align Docker Compose quick-start docs with required .env and loopback binding

Duansg:
## What's changed?

Follow-up to #4295 and #4257. The Docker Compose quick-start docs still describe the old flow, which no longer works:

- `docker compose up -d` fails immediately because `POSTGRES_PASSWORD` is now required, but the docs never mention `.env`.
- The docs say to open `http://ip:1157/`, but all published ports now bind to `127.0.0.1` by default.
- `.env.example` is not included in the `docker-compose.tar.gz` assembly, so users of the release package have nothing to copy.

Changes:

- `docker-compose-deploy.md` (en / zh-cn): add the `cp .env.example .env` step, switch the access URL to `localhost:1157`, and add a note on `HERTZBEAT_BIND_ADDRESS` / `HERTZBEAT_OTLP_BIND_ADDRESS` for remote access.
- `assembly-docker-compose.xml`: ship `.env.example` with every variant.
- `hertzbeat-postgresql-victoria-metrics/README*.md`: add the `.env` step and the same "Listener scope" section that #4295 added to the GreptimeDB variant; fix the title, which said IoTDB.
- zh-cn `upgrade.md`: sync the loopback-binding notes already present in the English version.

Verified with `markdownlint-cli2` (0 errors on the CI-linted files) and `script/ci/check-quickstart-compose.sh`.

## Checklist

- [ ] I have added the necessary unit tests and all cases have passed.

## Add or update API

### DIFF
--- a/home/docs/start/docker-compose-deploy.md
+++ b/home/docs/start/docker-compose-deploy.md
@@ -39,6 +39,15 @@ Run the `docker compose version` command to check if you have a Docker Compose e
    cd hertzbeat-postgresql-victoria-metrics
    ```
 
+   - Create the `.env` file and set the database password
+
+   > `POSTGRES_PASSWORD` is required. `docker compose up` refuses to start without it.
+
+   ```shell
+   cp .env.example .env
+   # Edit .env and set POSTGRES_PASSWORD to a strong password of your own
+   ```
+
    - One-click start
 
    > Run script in `hertzbeat-postgresql-victoria-metrics` directory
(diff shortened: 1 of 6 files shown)


zqr10159 (changes requested):
Thanks for aligning the quick-start documentation and release assembly. I verified the exact current head and found two items to address before approval:

1. The newly added English and Chinese deployment notes run `docker compose config`, but then tell users to restart with `docker-compose up -d`. The documented prerequisite checks the Compose v2 plugin, and a v2-only installation is not guaranteed to provide the legacy `docker-compose` executable. Please use `docker compose up -d` consistently in both new notes.
2. `script/docker-compose/hertzbeat-postgresql-victoria-metrics/README_CN.md` line 1 contains trailing whitespace, so the PR range fails `git diff --check`. Please remove it.

The substantive configuration and packaging behavior otherwise looks correct: all five Compose variants rendered successfully with the loopback defaults and independent bind overrides, the Java 25 startup/release package build passed, and the generated Docker Compose archive contains all five `.env.example` files with the expected fields.

Duansg:
@zqr10159  Thanks for the careful review, both points are fixed in 30de922fef.

1. **`docker compose up -d` consistency** — you're right, the new notes mixed the v2 plugin form (`docker compose config`) with the legacy standalone binary in the same sentence, which contradicts the documented `docker compose version` prerequisite. Both new notes now use `docker compose up -d`:
   - `home/docs/start/docker-compose-deploy.md`
   - `home/i18n/zh-cn/docusaurus-plugin-content-docs/current/start/docker-compose-deploy.md`

2. **Trailing whitespace** — removed from line 1 of `script/docker-compose/hertzbeat-postgresql-victoria-metrics/README_CN.md`. `git diff --check` is now clean over the full PR range.

Re-verified after the fix: `markdownlint-cli2` reports 0 errors across the CI-linted globs.

zqr10159 (approved):
Re-reviewed the remediation on the current head. The Compose v2 command is now consistent in both deployment notes, the trailing whitespace is removed, the full PR range passes `git diff --check`, all five Compose variants render successfully, and all current GitHub checks are green. No remaining blocking findings.

PR #4132, opened 2026-04-29
[feat] add ntfy as a new alert notification channel

Zmjjeff7:
### What is changed and how it works?

This PR adds [ntfy](https://ntfy.sh) as a new alert notification channel
(type=15) to HertzBeat. ntfy is a popular open-source HTTP-based
push-notification service that is widely used in self-hosting communities,
and it works well for receiving HertzBeat alerts on mobile devices and
desktops without depending on third-party platforms.

#### Backend changes
- New handler `NtfyAlertNotifyHandlerImpl` in `hertzbeat-alerter`,
  extending `AbstractAlertNotifyHandlerImpl` with type id `15`
- Three new fields on `NoticeReceiver`: `ntfyServerUrl`, `ntfyTopic`,
  `ntfyToken`
- New configurable property `alerter.ntfy-default-server-url`
  (default: `https://ntfy.sh`)

#### Feature highlights
- **Priority mapping** — alert severity is mapped to ntfy priority levels
  (`critical`=5, `warning`=4, `info`/unknown=3, resolved alerts=2)
- **Emoji tags** — status- and severity-aware tags
  (`rotating_light` + `skull` for critical firing, `warning` for warning,
  `white_check_mark` for resolved); the alertname is also appended as a
  plain-text tag
- **Bearer token auth** — supports self-hosted ntfy instances with
  access-control tokens
- **Click action** — notifications carry a click URL pointing to the
  HertzBeat console
- **Markdown body** — alert content is rendered in Markdown for
  better readability on ntfy clients

#### Frontend changes
- Added ntfy option to the notification receiver type dropdown
- Three new form fields: server URL (optional, default
  `https://ntfy.sh`), topic (required), access token (optional)
- Display tag in the receiver list table
- i18n translations for en-US, zh-CN, zh-TW, ja-JP, pt-BR

#### Tests
Added `NtfyAlertNotifyHandlerImplTest` with 16 test cases covering:
- Successful send / non-2xx response / network errors
- URL building (custom server, default server fallback,
  trailing-slash handling)
- Bearer token header injection
- Priority mapping for all severity levels and resolved state
- Tag generation for critical/warning/info/resolved cases
- Header presence and values (Priority, Tags, Click, Markdown)

### Checklist

- [x] I have read the [Contribution Guidelines](https://github.com/apache/hertzbeat/blob/master/CONTRIBUTING.md)
- [x] I have self-reviewed my code
- [x] I have added unit tests covering the new feature
- [x] My change follows the existing alert handler conventions
[...]

PR #4364, opened 2026-09-04
[fix] add missing ntfy default notice template

Duansg:
## What's changed?

`NtfyAlertNotifyHandlerImpl` declares `type() == 15`, but `resources/templates/` has no `15-*` file. `NoticeConfigServiceImpl` builds `PRESET_TEMPLATE` by scanning that directory, so `getDefaultNoticeTemplateByType((byte) 15)` returns null and `AlertNoticeDispatch#sendNoticeMsg` throws `NullPointerException` for every ntfy receiver that is not bound to a custom template — the notification is dropped. The "send test message" button fails the same way, since it also passes a null template.

Regression from #4132, which added the channel without its preset template.

- Add `15-NtfyTemplate.txt`, identical to the other markdown channels (the handler already sends `Markdown: yes`).
- Add `PresetNoticeTemplateCoverageTest` asserting the general invariant: every `@Component` `AlertNotifyHandler` other than SMS (which `AlertNoticeDispatch` exempts) has a preset template. It lives in `hertzbeat-manager` because that is the only module where both the template resources and the handler classes are on the same classpath. Verified it fails with `[NtfyAlertNotifyHandlerImpl(type=15)]` when the new template is removed.

`WeChatAlertNotifyHandlerImpl` (type 3) also has no template, but it carries no `@Component` and is never registered, so it is out of scope here and excluded by the test.

## Checklist

## Add or update API

### DIFF
--- a/hertzbeat-manager/src/main/resources/templates/15-NtfyTemplate.txt
+++ b/hertzbeat-manager/src/main/resources/templates/15-NtfyTemplate.txt
@@ -1,0 +1,47 @@
+## 🔔 HertzBeat Alert Notification
+
+### Alert Summary
+> - Status: ${status}
+<#if commonLabels??>
+<#if commonLabels.severity??>
+> - Severity: ${commonLabels.severity?switch("critical", "❤️ Critical", "warning", "💛 Warning", "info", "💚 Info", "Unknown")}
+</#if>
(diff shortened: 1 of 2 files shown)


yuluo-yx (approved):
LGTM

PR #4271, opened 2026-07-29
[fix] add missing rbac rules for the remaining unruled routes

Duansg:
## What's changed?

Add role constraints to the routes that had no rule (query, logs, alert reporting, labels, metrics, etc.), aligning them with their sibling routes.

## Checklist

- [x]  I have read the [Contributing Guide](https://hertzbeat.apache.org/docs/community/code_style_and_quality_guide)
- [ ]  I have written the necessary doc or comment.
- [x]  I have added the necessary unit tests and all cases have passed.

## Add or update API

- [ ] I have added the necessary [e2e tests](https://github.com/apache/hertzbeat/tree/master/e2e) and all cases have passed.

PR #4367, opened 2026-09-07
[fix] sync sureness.yml in release artifacts with the packaged rules

Duansg:
## What's changed?

#4257 and #4271 added the `/api/plugin/**` and `/api/account/token*` admin rules to `hertzbeat-startup/src/main/resources/sureness.yml`, but the six copies we actually ship were never updated:

- `script/sureness.yml` — the file `docs/start/docker-deploy.md` tells `docker run` users to download and mount over the one in the image.
- `conf/sureness.yml` in all five Docker Compose variants — compose mounts it as `./conf/sureness.yml:/opt/hertzbeat/config/sureness.yml`, overriding the image copy.

Sureness lets a path with no matching rule through for any authenticated role, so on both of those deployments **guest can still upload plugins and issue or revoke API tokens** — the two fixes above never took effect on the released artifacts. Four of the compose copies were also missing the `/observability/**`, `/metrics/**` and `/trace/**` static-resource exclusions.

This PR aligns all six copies with `hertzbeat-startup/src/main/resources/sureness.yml`; they now `diff` clean against it. Those omissions were the only differences — the `sureness.auths` account section is unchanged.

## Checklist

## Add or update API

### DIFF
--- a/script/docker-compose/hertzbeat-mysql-iotdb/conf/sureness.yml
+++ b/script/docker-compose/hertzbeat-mysql-iotdb/conf/sureness.yml
@@ -66,6 +66,10 @@ resourceRole:
   - /api/collector/**===post===[admin,user]
   - /api/collector/**===put===[admin,user]
   - /api/collector/**===delete===[admin]
+  - /api/plugin/**===get===[admin]
+  - /api/plugin/**===post===[admin]
+  - /api/plugin/**===put===[admin]
+  - /api/plugin/**===delete===[admin]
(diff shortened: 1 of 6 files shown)


zqr10159 (approved):
Reviewed the authorization-rule synchronization and release packaging path. The deployable sureness.yml copies match the canonical runtime configuration, the focused authorization/configuration tests pass, and the generated release artifacts contain the synchronized rules. No blocking issues found.

PR #4016, opened 2026-02-06
feat: SOP-driven AI workflows with skills, scheduling, and i18n support

TJxiaobao:
Summary
This PR delivers a product-ready AI workflow system for HertzBeat. It introduces SOP-based Skills (YAML-defined workflows) that combine tool calls and LLM reasoning, enabling automated diagnostics and reporting. It also adds scheduled execution with cron and conversation push-back, plus full i18n support for outputs in both interactive and background contexts.

Key Capabilities
SOP/Skills workflow engine with YAML definitions and unified output formats
Tool + LLM step executors to support automated diagnostics and analysis
Skill registry and auto tool discovery for extensibility
Scheduled SOP execution with cron, persistence, and conversation push results
Dynamic system prompt improvements and i18n coverage for AI workflows

Change List

SOP Engine: SopEngine, executors, models, registry, YAML loader
Skills: [daily_inspection.yml](app://-/index.html#), [mysql_slow_query_diagnosis.yml](app://-/index.html#)
Tools: ToolRegistry, SkillTools, ScheduleTools
Scheduling: schedule DAO/service/executor + cron support + persistence
i18n: SOP messages, background threads, dynamic language support
Prompt: injected dynamic skills and scheduling support

Demo
image
image
image
image
image
image

PR #4368, opened 2026-09-07
[fix] consolidate unreleased 1.9.0 migrations into V190 and fix PG/MySQL startup

Duansg:
## What's changed?

The unreleased `V182__scope_sop_schedule_owners.sql` fails to apply on PostgreSQL and MySQL, so the manager cannot start on either database.

- PostgreSQL: `SET enabled = 0` fails with `42804 column "enabled" is of type boolean but expression is of type integer`.
- MySQL: `CREATE INDEX idx_schedule_creator_conversation` fails with `1061 Duplicate key name`.

V183 is never reached. CI only ran H2, which accepts both statements — hence the green builds.

Both failures come from the same place. `FlywayConfiguration` delays Flyway until **after** Hibernate `ddl-auto` (`@DependsOn("entityManagerFactory")`), so `hzb_sop_schedule` is always created by Hibernate from `SopSchedule`, not by the migration. The column is therefore `boolean` / `bit(1)` rather than the `SMALLINT` / `TINYINT` the script declares, and `idx_schedule_creator_conversation` already exists because `@Index` declares it.

Following @zqr10159's review, this also fixes the migration layout. 1.8.0 ends at `V180` (tag `1.8.0`, 1b51c19c97), so V181/V182/V183 — all merged after 1.8.0 — are 1.9.0 content that drifted from the `V<major><minor><patch>` convention. They are replaced by one `V190__update_column.sql` per dialect.

## Changes

- Consolidate the unreleased migrations into a single `V190__update_column.sql` for h2, mysql and postgresql, in the order: create `hzb_sop_schedule` → disable ownerless schedules → enlarge `hzb_alert_define.expr`.
- Use the `FALSE` literal instead of `0`. On MySQL `FALSE` and `0` are the same value, so all three scripts stay portable.
- Declare `enabled BOOLEAN DEFAULT TRUE` in all three dialects, matching `SopSchedule.enabled`. Without this the script is only correct under the current Hibernate-first ordering: let the migration create the table and PostgreSQL gets `SMALLINT`, after which `SET enabled = FALSE` fails with the mirror-image `42804`.
- Drop the `CREATE INDEX` for `idx_schedule_creator_conversation`; `@Index` on the entity already owns it on both the fresh-install and the 1.8.x upgrade path.
- Keep the MySQL `DELIMITER` block last and precede it with `DROP PROCEDURE IF EXISTS`, so the routine body cannot swallow the plain statements above it and a re-run leaves nothing behind.
- Drop the trailing `COMMIT` from the old scripts and leave the transaction to Flyway.

## Migration regression, run by CI

`hertzbeat-startup/src/test/java/org/apache/hertzbeat/startup/migration/` — picked up by the existing `backend-build` job, which already runs `-Dmaven.test.skip=false` on a runner with Docker.

`MigrationInventoryTest` (no Docker) asserts every dialect ends the chain with V180 followed by V190, that V181/V182/V183 are gone, that the three dialects ship the same versions each exactly once, and that no stray file sits in a migration directory.

`AbstractMigrationUpgradeTest` drives the production startup order for each dialect: Hibernate `ddl-auto` builds the schema from the scanned entity model — same `@EntityScan` package and the same naming strategies Spring Boot applies — and only then does Flyway run, the way `FlywayConfiguration` orders them. The database is first taken to the last released **V180** state, so what the test exercises is the upgrade a 1.8.0 user performs. Flyway is configured by reading `spring.flyway` out of the shipped `application.yml`, not from a copy of it, so the test cannot drift from what the application runs.

Each dialect runs two scenarios:

| Scenario | What it models |
|---|---|
| schedule table owned by Hibernate | production: `CREATE TABLE IF NOT EXISTS` is a no-op and the script writes to Hibernate's column |
| schedule table created by the migration | the ordering that gave PostgreSQL `SMALLINT`, after which the script's own `SET enabled = FALSE` failed with 42804 |

Asserted on every combination: an ownerless schedule (`NULL` and `'   '`) ends disabled while `creator = 'alice'` stays enabled; `enabled` is a boolean in that dialect and `expr` is `CHARACTER LARGE OBJECT` / `LONGTEXT` / `TEXT`; a row seeded before the upgrade survives it and a 100 000 character `expr` round-trips afterwards; every required index exists exactly once, compared by column signature so a duplicate under another name is caught; no stored routine is left behind; the history holds `190` once and none of `181/182/183`; and a second startup applies nothing (`migrationsExecuted == 0`) and validates clean.

## Verification

`mvn -pl hertzbeat-startup,hertzbeat-manager test` on JDK 25, against real **MySQL 8.4** and **PostgreSQL 17** containers plus **H2 2.2.224** — 10 tests, all green.

The tests were mutation-checked: reverting each fix makes them fail with the original errors.

| Reverted fix | Failure |
|---|---|
| `enabled BOOLEAN` → `SMALLINT`, `FALSE` → `0` (PostgreSQL) | `V190 failed — SQL State 42804 — column "enabled" is of type boolean but expression is of type integer` |
| restore `CREATE INDEX idx_schedule_creator_conversation` (MySQL) | `V190 failed — SQL State 42000, Error Code 1061 — Duplicate key name 'idx_schedule_creator_conversation'` |
| drop the ownerless-schedule update (H2) | `a schedule without an owner must be disabled` |
| recreate the Hibernate-owned index (H2) | `unexpected indexes on hzb_sop_schedule … expected: <2> but was: <3>` |

One note surfaced by running this on a real server rather than reasoning about it: on MySQL the two paths produce different spellings of the same boolean — Hibernate declares `bit(1)`, the migration's `BOOLEAN` lands as `tinyint(1)`. Both hold `TRUE`/`FALSE` identically, and the test accepts either for MySQL while pinning PostgreSQL to `bool` so a `SMALLINT` regression is still caught.

Also confirmed end to end by running the 1.9.0 package on JDK 25 against PostgreSQL: before, startup aborts at V182; after, Flyway reports the schema at `v190`.

`mysql-connector-j` is added `test`-scoped to `hertzbeat-startup`: it is `provided`-scoped in `hertzbeat-manager` so it never reaches the test classpath, the version comes from the root `dependencyManagement`, and nothing `test`-scoped is packaged into the distribution. The PostgreSQL driver is already on that classpath and was not added.

Checklist

- [x] I have read the Contributing Guide

Add or update API

- [ ] I have added the necessary e2e tests and all cases have passed.


zqr10159 (commented):
Thanks for identifying and verifying the PostgreSQL/MySQL failures. The `FALSE` literal and removal of the duplicate Hibernate-managed index are correct, but please revise the migration layout before this is approved.

The release boundary should be the last official release, 1.8.0, whose migration chain ends at V180. RC1 and development snapshots are not compatibility targets for the final release. Therefore the migrations added after 1.8.0 should reflect the actual target release, 1.9.0:

- remove the unreleased V181, V182, and V183 files for H2, MySQL, and PostgreSQL;
- create one dialect-specific `V190__...sql` migration for each database;
- consolidate the effective V181-V183 behavior into V190: the SOP schedule schema changes, disabling schedules without a valid creator, and enlarging the alert expression column;
- retain this PR’s portable `FALSE` update and do not recreate `idx_schedule_creator_conversation`, because Hibernate already owns that index;
- add a migration inventory/upgrade regression that starts from the last official V180 state and verifies V190 on H2, MySQL, and PostgreSQL, including the resulting row state, column type, and indexes.

Databases created from RC1 or intermediate master snapshots may be reset or repaired manually; they should not force the final 1.9.0 migration numbering. I am leaving this as a non-approval review until the migration files are consolidated.

Duansg:
@zqr10159 Thanks — agreed on the numbering, and done.

Confirmed the boundary first: tag `1.8.0` (1b51c19c97) ends at `V180`, so V181 (#4016), V182 (#4280) and V183 (#4330) are all 1.9.0 content. They're now one `V190__update_column.sql` per dialect: create table → disable ownerless schedules → enlarge `expr`, keeping the portable `FALSE` and leaving `idx_schedule_creator_conversation` to `@Index`.

Two things worth flagging from the merge:

- **`enabled` is now `BOOLEAN` in all three dialects.** `FlywayConfiguration` delays Flyway until after Hibernate `ddl-auto`, so the table is always created by Hibernate and the migration's `CREATE TABLE` never runs — that's the root cause of the `42804`. Under the conventional Flyway-first ordering, PostgreSQL would create `SMALLINT` and `SET enabled = FALSE` would fail with the mirror-image error. `BOOLEAN` makes the script correct either way and matches `SopSchedule.enabled`.
- **MySQL's `DELIMITER` block is kept last**, preceded by `DROP PROCEDURE IF EXISTS`, so the routine body can't swallow the statements above it and a re-run stays clean.

Verified with `flyway-core` 10.11.1 (Flyway's own parser, not a SQL client) on H2 2.2.224, MySQL 8.4 and PostgreSQL 17, both execution orders, twice per run — all pass and are idempotent. The same harness on the old scripts reproduces `42804` and `1061`. Details in the PR description.

zqr10159 (commented):
The V190 consolidation is now structurally in the right direction, but two items still need to be addressed before approval.

**1. Remove the discarded migration-number narrative.**

The three V190 files still say that they consolidate `V181/V182/V183`. Those identifiers no longer exist in the release migration chain and should not remain in the final source. Please keep only release-facing wording such as `Schema changes for release 1.9.0.` The RC/snapshot history cleanup section in the PR description is also unnecessary because RC and intermediate master databases are not compatibility targets. `FlywayConfigurationTest` still uses `version 181` in its mock error text; please update it to V190 or make the example version-neutral.

**2. The migration verification is not yet sufficient as a repository merge gate.**

The latest diff changes only SQL files; it adds no test source or CI wiring. The external Flyway/Testcontainers harness described in the PR is useful evidence, but it cannot be reproduced by another maintainer from this repository and will not protect the migration from regression. The existing backend E2E jobs do not execute these startup migrations against all three database dialects.

Please commit a focused migration test that CI runs and that:

- asserts that each dialect has V180 followed by V190, with no V181/V182/V183 files;
- uses the repository Flyway 10.11.1 configuration to migrate the last official V180 state to V190 on H2, MySQL, and PostgreSQL;
- follows the production Hibernate-first/Flyway-second startup order (and tests Flyway-first too if that remains a stated supported behavior);
- verifies ownerless schedules are disabled, a valid owner remains enabled, `expr` has the expected dialect type, required indexes exist exactly once, and the MySQL helper procedure is removed;
- performs a second application/Flyway startup and proves it is a clean no-op without validation or duplicate-object failures.

Once these stale references are removed and the real migration proof is committed and running in CI, this should be ready for another review.

zqr10159:
Clarification on the test request: there is no requirement to commit the separate local harness itself. The repository’s existing Maven E2E framework can cover this properly.

Please add a focused `hertzbeat-startup-e2e` module under `hertzbeat-e2e`, add it to `hertzbeat-e2e/pom.xml`, and include it in the `backend-maven-e2e` module list in `.github/workflows/backend-build-test.yml`. The E2E parent already provides Testcontainers, and the repository already uses the MySQL/PostgreSQL Testcontainers modules and JDBC drivers.

The test should use in-memory H2 plus MySQL and PostgreSQL containers, run the repository Flyway configuration from a V180 baseline to V190, and verify the resulting data/schema state and a clean second startup. In particular, assert the ownerless/owned schedule behavior, final `expr` type, index uniqueness, MySQL procedure cleanup, and Flyway current version V190. A small migration-inventory assertion should also ensure the three dialect directories contain V190 rather than the discarded intermediate migration files.

This gives us reproducible CI evidence without promoting a one-off local test tool into the codebase. The earlier request to remove the old migration-number narrative from the V190 files and PR description still applies.

Duansg:
@zqr10159 Both points are addressed in 32895b0aed.

**1. Stale identifiers removed.** The three V190 headers now read only `Schema changes for release 1.9.0.` plus a note that every statement is re-runnable; `V181/V182/V183` are gone from the scripts, and so are the issue references. `FlywayConfigurationTest` no longer mentions `version 181`. The RC/snapshot history-cleanup section is dropped from the PR description.

**2. The proof is committed and CI runs it.** `hertzbeat-startup/src/test/.../migration/` — picked up by the existing `backend-build` job, which already runs `-Dmaven.test.skip=false` on a Docker-capable runner, so it is a merge gate and not a job that has to be wired up.

- `MigrationInventoryTest` (no Docker): every dialect ends V180 → V190, no V181/V182/V183, all three dialects ship the same versions each exactly once, no stray file in a migration directory.
- `AbstractMigrationUpgradeTest` + one subclass per dialect: Hibernate `ddl-auto` builds the schema from the scanned entity model — same `@EntityScan` package and the same naming strategies Spring Boot applies — and Flyway runs after it, the way `FlywayConfiguration` orders them. The database is taken to the last released **V180** state first, so the scenario is the upgrade a 1.8.0 user performs, not a synthetic one. Flyway is configured by reading `spring.flyway` out of the shipped `application.yml` rather than a copy, so the regression cannot drift from what the application runs.

On Flyway-first: it is not a supported ordering and the test does not pretend otherwise. V160 and V180 shipped assuming Hibernate-created tables — on an empty database V180's first statement is an unguarded `UPDATE HZB_ALERT_DEFINE`, which cannot work. What is worth covering is the part of that ordering V190 must survive, so the second scenario per dialect drops `hzb_sop_schedule` after V180 and lets the migration create it. That is exactly the path that gave PostgreSQL `SMALLINT` and then broke on the script's own `SET enabled = FALSE`.

Asserted on every combination: `NULL` and `'   '` owners end disabled while `creator = 'alice'` stays enabled; `enabled` is a boolean in that dialect and `expr` is `CHARACTER LARGE OBJECT` / `LONGTEXT` / `TEXT`; a row seeded before the upgrade survives it and a 100 000 character `expr` round-trips after it; each required index exists exactly once, compared by column signature so a duplicate under a different name is caught; no stored routine is left behind; the history holds `190` once and none of `181/182/183`; a second startup applies nothing and validates clean.

**Verified**, not just written: JDK 25 against real MySQL 8.4 and PostgreSQL 17 containers plus H2 2.2.224 — 10 tests green. Then mutation-checked by reverting each fix:

| Reverted | Failure |
|---|---|
| `BOOLEAN`→`SMALLINT`, `FALSE`→`0` (PG) | `42804 column "enabled" is of type boolean but expression is of type integer` |
| restore `CREATE INDEX idx_schedule_creator_conversation` (MySQL) | `42000 / 1061 Duplicate key name 'idx_schedule_creator_conversation'` |
| drop the ownerless-schedule update (H2) | `a schedule without an owner must be disabled` |
| recreate the Hibernate-owned index (H2) | `unexpected indexes … expected: <2> but was: <3>` |

Two things running this against real servers turned up that reasoning alone had not. On MySQL, Hibernate declares the boolean as `bit(1)` while the migration's `BOOLEAN` lands as `tinyint(1)` — same values, two spellings — so the test accepts either there while pinning PostgreSQL to `bool`, which keeps the `SMALLINT` regression detectable. And MySQL Connector/J leaves `nullCatalogMeansCurrent` false, so an unqualified `DatabaseMetaData` lookup silently matched a same-named table in another test's database; the lookup is now pinned to the connection's catalog and schema.

One dependency note, since it is the only build change: `mysql-connector-j` is added `test`-scoped to `hertzbeat-startup`, because it is `provided`-scoped in `hertzbeat-manager` and so never reaches the test classpath. The version comes from the root `dependencyManagement`, it is already a dependency of four other modules, and nothing `test`-scoped is packaged into the distribution. The PostgreSQL driver was already on that classpath, so it is not added.

zqr10159 (approved):

PR #4389, opened 2026-09-20
fix(monitor): complete the missing help links in the monitoring template

P-Peaceful:
## What's changed?

- Added missing `helpLink` entries to 11 monitoring templates:
  - Dahua
  - DNS
  - Hikvision ISAPI
  - Memcached
  - Modbus
  - NTP
  - Registry
  - S7
  - SMTP
  - Uniview
  - WebSocket
- Added localized help links for both Chinese and English documentation.
- The S7 monitoring documentation is not available yet. It will be completed and submitted in a follow-up PR.
- This allows users to access the corresponding monitoring documentation directly from the monitoring configuration interface.

## Checklist

## Add or update API

## Testing

Not run. This is a configuration-only change that adds documentation links to monitoring templates.

## Follow-up

The S7 monitoring help documentation is currently unavailable. I will add the missing S7 documentation in a subsequent PR.


hertzbeat-manager/src/main/resources/define/app-s7.yml
```diff
@@ -28,6 +28,9 @@ help:
   en-US: HertzBeat collects metrics related to maintaining registers and coils for services that support s7 protocol
   zh-TW: HertzBeat對支持s7協定的服務進行（保持寄存器和線圈）相關名額進行採集
   ja-JP: HertzBeatはs7プロトコルをサポートするサービスのメトリクスを監視します。
+helpLink:
```
utafrali: The S7 help documentation doesn't exist yet (acknowledged in the PR description), so both URLs here will return 404 until the follow-up PR lands. A dead link is a worse user experience than no link at all. Consider one of these approaches:
1. Drop the `helpLink` block from this file now and add it in the same PR that ships the S7 docs.
2. Or, if the docs page already exists in draft/stub form, verify the URLs resolve before merging.

utafrali (approved):
Straightforward config-only change that adds missing `helpLink` entries to 11 monitoring templates, consistently following the existing `zh-CN`/`en-US`-only pattern seen in templates like `app-linux.yml`. The only concern is the S7 link pointing to documentation that doesn't exist yet.

hertzbeat-manager/src/main/resources/define/app-s7.yml
P-Peaceful: ok, i willed delete this `helpLinke`

